PR #1, opened 2023-08-24
Scaffolding for SHACL validation (ontology and examples)

avillar:


rob-metalinkage (approved):
Makes no changes to normative content - adds subdirectories to ssn which is fine for next refactoring conversation-  will trigger generation of validation examples to feed into this conversation

dr-shorthair (approved):
This will allow us to move forward on continuous testing of ontology + examples

PR #91, opened 2023-09-04
Move owl:inverseOf axioms to SSN graph

### LINKED ISSUE #90: Which graph do owl:inverseOf axioms belong to?
Following on from #88 and #89:

Most of the axiomatization in SSN/SOSA is housed in the SSN graph. 
All sub-class, sub-property, restrictions, property-chain-axioms and most inverseOf are in the SSN graph. 
The SOSA graph only has axioms for `rdf:type` and some `owl:inverseOf`. 

Should the remaining `owl:inverseOf` axioms be moved to the SSN graph?

dr-shorthair:
Fixes #90


rob-metalinkage (commented):
To confirm that we should not risk backwards incompatibility with this change

PR #92, opened 2023-09-04
Mirror ssn resources in sosa namespace

### LINKED ISSUE #48: Should we merge SSN and SOSA into a single namespace?
The idea would be to define all classes in SOSA, including Deployment, System, etc. that only exist in the SSN namespace as of today.

Even if we do that, we could keep the existing classes in the SSN namespace for backwards compatibility and just map them as equivalent to the new ones we add to SOSA namespace.

What do we do with constraints that are only defined in SSN in the current version?

dr-shorthair:
All the `ssn:` classes and properties are mirrored into the `sosa:` namespace but remain in the SSN graph. 
The existing `ssn:` resources have `owl:equivalentClass` or `owl:equivalentProperty` axioms to the new resources. 

Fixes #48


rob-metalinkage (dismissed):
Refactor to remove commit removing inverseOf

rob-metalinkage (commented):
Should sosa declare equivalence - or only ssn?  I guess a ssn+ sosa enclosure works and sosa alone needs to know nothing about ssn so this is safe and sensible,

dr-shorthair:
> Refactor to remove commit removing inverseOf

The need for this `inverseOf` property depends somewhat on the final resolution of the `madeBySensor` vs `madeByObserver` issue. Are they `equivalentPropertys`? 

I will re-instate it for now, so we can move forward.

dr-shorthair:
Hmm. If both `madeBySensor` and `madeByObserver` are `inverseOf` the same property, then I think this entails they are `equivalentProperty`.

dr-shorthair:
> Should sosa declare equivalence - or only ssn? 

Unclear which resource should have the `owl:equivalentX` axiom attached. 
The original `ssn:` resource? The new `sosa:` resource? Both? 

For the time being they sit on the `sosa:` resource so the original resources are untouched internally.

dr-shorthair:
Close and re-create as merge into `gh-pages`.

PR #97, opened 2023-09-21
Mirror ssn resources in sosa

### LINKED ISSUE #48: Should we merge SSN and SOSA into a single namespace?

dr-shorthair:
- All SSN classes and properties cloned into SOSA namespace and graph, following existing axiomatization and annotation rules applicable to each graph (file)
- `owl:equivalentClass` and `owl:equivalentProperty` axioms added to both sides, in SSN graph

Fixes #48


rob-metalinkage (commented):
Looks good - but are we replicating stuff Alex Robin did ?

dr-shorthair:
Did Alex create new classes/properties in the SSN namespace?

rob-metalinkage (approved):
Discussed in meeting 21/09/2023 - happy this consolidates @AlexRobins input ready for issue processing.

sgrellet (approved):
Aligned with discussion in meeting 21/09/2023

PR #100, opened 2023-09-23
Remove `sosa:Observer` retain `sosa:Sensor`

### LINKED ISSUE #29: Change prefLabel of sosa:Sensor to 'Observer'
The `Sensor` class is an abstract concept, whose extension includes all agents, devices, software etc that is capable of implementing a procedure in order to evaluate the state of some property. However, the name 'Sensor' is usually only associated with a subset of this conceptual space - i.e. mechanical and electronic sensors. While there is no need to change the URI, it would be helpful and informative to modify the annotation so that users get a better immediate indication of the scope of this class. 

Note that the revision of ISO 19156 currently underway in ISO/TC 211 is adopting many of the innovations from SSN/SOSA/SSN-ext into the UML version of this model. They have already decided to name the corresponding class 'Observer' instead of Sensor.

### LINKED ISSUE #43: Should Observer and Host classes be equivalent to or superclass of Sensor and Platform?
In PR w3c/sdw#1402, `Observer` and `Host` are super classes of their `Sensor` and `Platform` counterparts. They are also kept more abstract because they are not sub-classes of `System`.

[Link to specific commit](https://github.com/w3c/sdw/pull/1402/commits/638f6c462d11d406c5a093cd0b0d522cc161231a)

Alternatively, we could have `Observer sameAs Sensor`, and `Host sameAs Platform`.

dr-shorthair:
closes #29


rob-metalinkage (commented):
looks good - only thing is whether the retention of Sensor should explicitly refer to backwards compatibility concerns.

rob-metalinkage (approved):

oldskeptic:
I think this looks good.

oldskeptic (approved):
Looks good to go.

PR #119, opened 2023-11-17
Add `observationType` and `hasQualityMeasurement`

dr-shorthair:
Additional Observation properties mentioned in OMS

### DIFF
--- a/ssn/integrated/sosa-oms.ttl
+++ b/ssn/integrated/sosa-oms.ttl
@@ -12,7 +12,7 @@
 
 @prefix schema: <http://schema.org/> .
 
-@prefix ogc: <http://www.opengis.net/def/ont/modspec/> . 
+@prefix ogc-ms: <http://www.opengis.net/def/ont/modspec/> . 
 @prefix sosa: <http://www.w3.org/ns/sosa/> .
 @prefix sosa-oms: <http://www.w3.org/ns/sosa/oms/> .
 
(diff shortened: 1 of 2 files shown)


KathiSchleidt (approved):
Looks good!

maximelefrancois86 (changes requested):
Why is `observationType` in the SOSA namespace ? This object property is not defined in sosa.ttl  and not documented in the specification.

dr-shorthair:
It comes from OMS. 
I put it in `sosa:` ns in anticipation of it being added to sosa.ttl. 
See https://github.com/w3c/sdw-sosa-ssn/issues/132

dr-shorthair:
My suggestion is that we put all the new things in `sosa:` provisionally. 
We can then consider if we need more horizontal modules in due course, which might involve moving some terms into new namespaces. 

(We have a lot of branches and need to do some merging soon else it'll get horribly tangled.)

dr-shorthair:
Closes #124

ssn/integrated/sosa-oms.ttl
```diff
@@ -307,6 +300,12 @@ sosa:Observation ogc:implements <http://www.opengis.net/spec/om/3.0/req/obs-cpt/
     schema:rangeIncludes sosa:Observation ;
     rdfs:isDefinedBy sosa-oms: .
 
+  dqv:hasQualityMeasurement
+    ogc:implements <http://www.opengis.net/spec/om/3.0/req/obs-core/AbstractObservationCharacteristics/resultQuality-sem> ;
+    skos:definition "Information pertaining to the data quality of the result." ;
+    schema:domainIncludes osa:Observation ;
```
sgrellet: sosa:Observation instead of osa:Observation

I guess we'll discuss the range afterwards

sgrellet: the rest looks ok to me

sgrellet (approved):
fine with me

dr-shorthair:
> Why is `observationType` in the SOSA namespace ? This object property is not defined in sosa.ttl and not documented in the specification.

For clarity at this stage, I've moved all the terms that are defined in the SOSA-OMS graph into the `sosa-oms:` ns. 
We can consider the final destination in #132. 
@maximelefrancois86 can you approve now so we can merge?

maximelefrancois86 (approved):

PR #145, opened 2024-01-04
Retire O&Mv2 alignment

### LINKED ISSUE #128: Alignment of iso19156-om:OM_Observation.featureOfInterest with sosa:hasFeatureOfInterest
This is a very minor issue:

In section 6.3.5 Property Alignments, alignment to OM, it is stated that:

 iso19156-om:OM_Observation.featureOfInterest	equivalent property  sosa:hasFeatureOfInterest

Actually sosa:hasFeatureOfInterest can also be applied to Actuations. And depending on the resolution of issue #111 , there may be even more classes that it will apply to.

I suggest to change this alignment as follows: 

 iso19156-om:OM_Observation.featureOfInterest	sub property of   sosa:hasFeatureOfInterest

And see if the same reasoning applies to other alignments.

dr-shorthair:
See #128


sgrellet (approved):
As per 2024-01-04 webconf discussion

KathiSchleidt (approved):
Makes sense, alignment between O&M and OMS is in the OMS spec, no reason to map to an old version here

Issue #140, opened 2023-12-14
Re-draw figures - add new classes and properties


dr-shorthair:
See figure captions marked 'TODO'

dr-shorthair:
@maximelefrancois86 do you recall who did the diagrams in the 2017 edition? 
What tool did they use?

maximelefrancois86:
@rgcmme did the diagrams. Probably using draw.io / diagrams.net
I don't have the sources of these diagrams.

rgcmme:
Yes, I made the diagrams using Powerpoint.

The sources are already in the repo:

https://github.com/w3c/sdw-sosa-ssn/blob/gh-pages/ssn/SOSA-SSN%20ontologies%20v6.pptx

If I had to do them today I'd follow the Chowlk notation:

https://chowlk.linkeddata.es/

dr-shorthair:
> If I had to do them today I'd follow the Chowlk notation:

It appears that Chowlk diagram --> TTL is supported, but not the other direction.
We have the TTL, we want diagrams. 
Or am I misunderstanding?

rgcmme:
That's it, right now the only possible way is from diagram to OWL. I don't know of any tool that would make the opposite in a satisfactory way.

dr-shorthair:
Guess I better learn draw.io

dr-shorthair:
Potential replacement for Figure 1 - 
https://github.com/w3c/sdw-sosa-ssn/blob/183-clarify-ssn-ontology/ssn/images/SOSA-modules.png

![New Figure 1?](https://raw.githubusercontent.com/w3c/sdw-sosa-ssn/183-clarify-ssn-ontology/ssn/images/SOSA-modules.png)

PR #146, opened 2024-01-04
Various non-normative HTML fixes

### LINKED ISSUE #143: Format index.html source
The indentation in ssn/index.html is currently inconsistent, which makes editing more difficult. 
Various tools can fix this automatically (e.g. VSCode plugins). 
However, this trivial (non-normative) action will change a large number of lines in the file so will look like a big diff. 

Propose to do it once quickly when we have few branches.

dr-shorthair:
fixes #143

### DIFF
--- a/ssn/index.html
+++ b/ssn/index.html
@@ -6,8 +6,8 @@
   <title>Semantic Sensor Network Ontology</title>
   <script class="remove" src="https://www.w3.org/Tools/respec/respec-w3c-common"></script>
   <script class="remove" src="config.js"></script>
-  <script type="text/javascript" src="https://www.w3.org/2007/OWL/toggles.js"></script>
-  <script type="text/javascript">
+  <script src="https://www.w3.org/2007/OWL/toggles.js"></script>
+  <script>
 
     function hide_ssn(hide, element) {
       var scrollY = window.scrollY;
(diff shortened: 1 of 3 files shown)


KathiSchleidt (approved):

dr-shorthair:
Could one more person review so this can be merged?

ldesousa (changes requested):
@dr-shorthair This is a very large set of changes, since it is non-normative I didin't look into detail to the content, but a few points are worth considering:

1. There are many "... TODO ..." notes across the document. Is it on purpose? Eve if yes consider instead to open an issue pointing to the missing content.

2. It looks like some of the inner blue tables are missing cells (i.e. the background is not applied). Some examples below.

Selection_121
Selection_122

3. There are many broken links throughout the document. In Brave (Chromium) the HTML render marks those with a red wavy underline. A few examples below.

Selection_126
Selection_125
Selection_124
Selection_123

dr-shorthair:
Thank you @ldesousa 

1. The TODO comments are either
  - need to redraw figures - see #140 - but note that this will be deferred until the ontology is stable
  - need for examples for the Collection terms - see new issue #159
2. The inner blue tables show the axioms in the SSN graph. If you click the link `[Hide additional SSN axioms]` it will be hidden. I've added a note in the Modularization section drawing attention to this capability
3. I've fixed 7 broken links (internal cross-references). The red underlines associated with the show/hide buttons are an artefact of javascript and relative paths which should go away when the branch is merged. 

I also accepted the lint corrections from @sgrellet 

See https://raw.githack.com/w3c/sdw-sosa-ssn/fix-html/ssn/index.html#Specification for the revised state

ldesousa (approved):

PR #153, opened 2024-01-06
Update OBOE-SOSA mapping to use ObservationCollection

### LINKED ISSUE #151: Update OBOE alignment
to match SSN-ext as implemented in SOSA-update.

dr-shorthair:
Closes #151


ldesousa:
There is something I don't understand (therefore not blocking the merge): why are `oboe-core:Characteristic` and `oboe-core:Measurement` mapped with `owl:equivalentClass`, while `oboe-core:Observation` and `oboe-core:ObservationCollection` are mapped with `rdfs:subClassOf`? I understand the concept of the latter two being specialisations of `sosa:ObservationCollection`, I just wonder if there are other implications in using different predicates for the alignment.

dr-shorthair:
Thanks for looking at this carefully @ldesousa . 

The difference in the mappings was carefully considered. 

The mappings are best understood using set theory. `owl:equivalentClass` means that the class extension (membership) is the same [1], while `rdfs:subClassOf` means that the class extension is a subset. Don't confuse `owl:equivalentClass` with `owl:sameAs` as the latter implies that actual graphs can be merged, which is a different thing. 

OBOE provides a general model for observations, very comparable to the SOSA `Observation` and `ObservationCollection` structures.  

Looking at OBOE and SOSA, _every_ individual `oboe-core:Measurement` could be expressed as an individual `sosa:Observation`, _and vice versa_.  Ditto for `oboe-core:Characteristic` and `sosa:ObservableProperty`. Hence `oboe-core:Characteristic` and `oboe-core:Measurement` are `equivalentClass` to `sosa:ObservableProperty` and `sosa:Observation`, respectively.

In contrast, two classes from OBOE (`oboe-core:Observation` and `oboe-core:ObservationCollection`) are implemented by only one class from SOSA (`sosa:ObservationCollection`), with some restrictions on collection membership. i.e. while every `oboe-core:Observation` is also a `sosa:ObservationCollection`, and every `oboe-core:ObservationCollection` is also a `sosa:ObservationCollection`, only some `sosa:ObservationCollections` are `oboe-core:ObservationCollections` and only some `sosa:ObservationCollections` are `oboe-core:Observations`, and some are neither. Furthermore, the two OBOE classes are disjoint, ie there is no intersection between these. Thus the mapping is a subclass relationship. 

Finally, note that this mapping is labeled 'non-normative'. 

[1] https://www.w3.org/TR/owl-ref/#equivalentClass-def

ldesousa (approved):

dr-shorthair:
Could one more person review this PR so the changes can be merged

dr-shorthair:
Note that this is a straightforward alignment, originally documented in [SSN-ext](https://www.w3.org/TR/vocab-ssn-ext/#oboe). 
I've just re-used the figures from there. 

It was one of the things that led to the `ObservationCollection` proposal. 

Need one more review to merge.

dr-shorthair:
As usual, the browser view for this branch can be seen at https://raw.githack.com/w3c/sdw-sosa-ssn/OBOE-mapping/ssn/index.html

rob-metalinkage (approved):
Thanks for the detailed clarification - looks sound.

PR #155, opened 2024-01-06
Retire ssnx alignment

### LINKED ISSUE #154: Retire SSNX alignment module?
The 2017 version of the SSN Ontology spec has a (non-normative) of the alignment with the 2011 Ontology ('SSNX') that was the result of the previous W3C Incubator group - see https://www.w3.org/TR/vocab-ssn/#SSNX_Alignment  . This was considered important due to the highly-cited status of a paper describing the 2011 work https://www.sciencedirect.com/science/article/pii/S1570826812000571?via%3Dihub#section-cited-by . 

We have preserved all `sosa:` and `ssn:` terms from the 2017 version, so the [alignment file](https://www.w3.org/2017/01/ssn-ssnx/) is still OK. However, we have now deprecated use of the `ssn:` terms, and the link back to the 2011 ontology is of declining significance. 

Should we remove or reduce the scale of this section, maybe leaving behind a mention of the tabulated alignment in the 2017 edition of this spec and a link to the alignment file?

dr-shorthair:
Fixes #154


dr-shorthair:
Do we still need the SSNX alignment in the Rec document? It can always be recovered from the earlier edition if needed, as indicated in the paragraph I've left behind - see 
https://raw.githack.com/w3c/sdw-sosa-ssn/Retire-SSNX-alignment/ssn/index.html#SSNX_Alignment

rob-metalinkage (approved):
Looks reasonable to me.

sgrellet (approved):
Looks good to me.

PR #165, opened 2024-01-17
Clarify the role and cardinality of Actuation Result; add ActuationCollection

### LINKED ISSUE #112: What is the result of an actuation ?
An actuation can use a procedure, that is described in terms of input and output.

I can guess the output of an observation procedure is a description of the expected result.
I can guess the input of an actuation procedure is a description of the expected parameters of the actuation.

Now how does the 'input' of the procedure translates on the observation itself ?
And what is the result of an actuation ? is it a resulting property value ? a response code ? 

Wouldn't it be more appropriate to model the _argument_ of an actuation ?

Proposal: clarify input/output in SSN, and model the _arguments_ of an actuation (`sosa:hasArgument`)

dr-shorthair:
Closes #112 #174


ssn/integrated/sosa.ttl
```diff
@@ -452,11 +453,10 @@ sosa:Observation a rdfs:Class , owl:Class ;
     rdfs:isDefinedBy sosa: .
 
 
-
-sosa:Actuation a rdfs:Class , owl:Class ;
+sosa:Actuation a owl:Class ;
   rdfs:label "Actuation"@en ;
-  skos:definition "An Actuation carries out an (Actuation) Procedure to change the state of the world using an Actuator."@en ;
-  rdfs:comment "An Actuation carries out an (Actuation) Procedure to change the state of the world using an Actuator."@en ;
+  skos:definition "An Actuation carries out an (Actuation) Procedure to change the state of the world using an Actuator. hasResult indicates the intended value of the ActuatableProperty at the completion of the Actuation"@en ;
```
hylkevds: > at the completion of the Actuation

at some point in the future? When changing a thermostat It may take hours before the temperature has reached the desired value. At the completion of the Actuation the heating/cooling system is only _starting_ to change the temperature.

ssn/index.html
```diff
@@ -1513,9 +1513,14 @@ <h5><a href="#SOSAActuation">sosa:Actuation</a></h5>
           <p><strong>an OWL Class</strong></p>
 
           <em property="rdfs:label">Actuation</em> -
-          <span property="rdfs:comment skos:definition">An <a href="#SOSAActuation"><code>Actuation</code></a> carries
-            out an (Actuation) <a href="#SOSAProcedure"><code>Procedure</code></a> to change the state of the world
-            using an <a href="#SOSAActuator"><code>Actuator</code></a>.</span> <br>
+          <span property="skos:definition">An <a href="#SOSAActuation"><code>Actuation</code></a> carries out an
+            (Actuation) <a href="#SOSAProcedure"><code>Procedure</code></a> to change the state of the world using an <a
+              href="#SOSAActuator"><code>Actuator</code></a>. <a href="#SOSAhasResult"><code>hasResult</code></a>
+            indicates the intended value of the <a href="#SOSAActuatableProperty"><code>ActuatableProperty</code></a> at
```
hylkevds, line 1524: > the ActuatableProperty

What if there are multiple? How can we differentiate between them?

ssn/integrated/sosa.ttl
dr-shorthair: There are many complexities in specific actuation situations that are beyond the scope of this general framework. However, there is one common requirement across all actuations: the target state i.e. the intended value of the actuatable property. That is how I understood it. 

More complex scenarios must be modeled in applications that extend or specialize SSN/SOSA.

ssn/index.html
dr-shorthair: As [mentioned in the other thread](https://github.com/w3c/sdw-sosa-ssn/issues/112#issuecomment-1899103582) that was news to me. 
Clearly this is an issue that needs to be addressed. 

My preferred approach is to limit cardinality to 1, and push the question of correlating result and property vectors into future work.

alexrobin: I agree that we should keep the same pattern as Observation and set the cardinality to exactly ONE.

Although we have applications of SOSA/SSN where we use result and property vectors, I'm not sure this should be in the scope of the ontology. Applications can define their own mechanisms to handle this (as we do in connected systems).

dr-shorthair: Establishing ActuationCollection would provide a standard pathway for applications with multiple ActuatableProperties and Results. 
I'll have a go at that.

dr-shorthair: OK - I've added ActuationCollection in this branch

dr-shorthair:
Closes #174

dr-shorthair:
As usual, the browser view for this branch can be seen at https://raw.githack.com/w3c/sdw-sosa-ssn/actuation-result/ssn/index.html

dr-shorthair:
superseded by #178

Issue #105, opened 2023-10-23
Add `sosa:Observation` to the domain of `hasInput` 


dr-shorthair:
Some 'Observations' use previously generated information as input, in addition to or instead of sensor stimuli. This particularly applies to 'observations' where the procedure is some complex processing or interpretation algorithm. 

`Observation` should be added to the `schema:domainIncludes` for `hasInput` to allow specific input values to be indicated, in addition to any general inputs indicated in the procedure description.

dr-shorthair:
Or should it be renamed `hadInput` in this context?

maximelefrancois86:
Good idea !
Then what about `hasOutput`? How would it relate to `hasResult`?

maximelefrancois86:
> Or should it be renamed `hadInput` in this context?

I think this single-letter difference has the potential to generate great confusion among  users

hylkevds:
There are two different uses of `hasInput`:
- The `hasInput` of the `Procedure` points to a _parameter description_.
- The `hasInput` of the `ProcedureExecution` must point to a _parameter value_ **and** to the _parameter description_ of the `hasInput` of the Procedure that this value is for.

These are different object properties, so using different names would be better. How about `hasInputValue` for the second one?
I suppose that for the second one we'll still need a class.

dr-shorthair:
I agree.

dr-shorthair:
#173 introduces `hasInputValue` and `hasOutputValue`

dr-shorthair:
Closed by #173

PR #173, opened 2024-01-18
Content model for Execution (was ProcedureExecution)

### LINKED ISSUE #164: Specification of ProcedureExecution
Related #109 

While looking at #112 I'm pondering if `ProcedureExecution` is specified enough. 
Several of the properties of the subclasses `Observation` `Sampling` and `Actuation` are common to all (that's the point really) so  could/should be associated with `ProcedureExecution`. Probably with relaxed cardinalities. 

I suggest: 
- `sosa:usedProcedure only sosa:Procedure`
- `sosa:madeOnPlatform only sosa:Platform`
- `sosa:hasFeatureOfInterest only sosa:FeatureOfInterest`
- `sosa:hasUltimateFeatureOfInterest only sosa:FeatureOfInterest`
- `sosa:resultTime exactly 1 `

dr-shorthair:
Closes #164


dr-shorthair:
Should `hasInputValue` be added - see #105 ?

rob-metalinkage (commented):
A super  class has value for schema implementation - so agree.  Consider using Activity from the PROV alignment. 

Evidence of implementation can be supported by a common schema in the OGC PAI-SOSA implementation.

ldesousa:
After the explanation during today's conference call the role of this class became clear. However, I am concerned the name may generate some confusion. It is not as concise as the other main concepts and contains the name of an existing class (`Procedure`). Names like `Action` (OMD software) or `Activity` (from Prov ontology) to my mind would be more effective.

sgrellet:
agreed. 

during Jan 18th webconf the issue of name conflict between `ProcedureExecution `and `sosa:Procedure` was raised. 

Not the interest of this new class (which makes a lot of sense as it gives reality to an implictly common super class), just its naming. To avoid misunderstanding by users.

A suggestion was to name it Activity. And it would actually be interesting to consider link with prov:Activity (subtyping?)

dr-shorthair:
@maximelefrancois86 it seems that the name `ProcedureExecution` is causing some confusion. 
Would renaming to `sosa:Activity` (parallel to `prov:Activity`) cause any problems?

maximelefrancois86:
I'd be ok with a simpler name, but note many properties that would be attached to that class would be weird if attached to `Activity`:
- `madeBy`
- `hasFeatureOfInterest`
- `hasUltimateFeatureOfInterest`
- `hasResultTime`

Users may wonder why such a general concept like `Activity` needs to be made by a sensor/actuator, be about a feature of interest, and have exactly one result...

I thought the name `ProcedureExecution` would help to make clear the distinction with `Procedure`.

DUL and PROV-O alignments are both important and structuring.

PROV-O offers Activity as you pointed out

DUL offers:
> Event: Any physical, social, or mental process, event, or state.
> Action -> An Event with at least one Agent that isParticipantIn it, and that executes a Task that typically isDefinedIn a Plan, Workflow, Project, etc. ------> this perfectly aligns to our idea of what a procedure execution is 
> Process -> This is a placeholder for events that are considered in their evolution, or anyway not strictly dependent on agents, tasks, and plans.

hylkevds:
One advantage of the name ProcedureExecution is that it is quite clearly the Execution of a Procedure, and not some other, non-Procedure-related Activity.
How about just `Execution`? Since that name implies that there is something else that is being executed (a program or procedure).

maximelefrancois86:
I'd be happy with `Execution`

dr-shorthair:
I just added `hasInputValue` and `hasOutputValue` for use on `Executions`. 
But I have a feeling more documentation and justification or patterns is required. 
Can @hylkevds or @alexrobin or @maximelefrancois86 or @sgrellet step in here please? 

FWIW I'm somewhat uncertain of the `hasOutputValue` semantics and how it relates to `hasResult`.

dr-shorthair:
As usual, the browser view for this branch can be seen at https://raw.githack.com/w3c/sdw-sosa-ssn/ProcedureExecution-spec/ssn/index.html

hylkevds:
`hasInputValue` and `hasOutputValue` will need two properties: 
- One that points to what the `hasInput` / `hasOutput` of the Procedure points to. How about `forInput` / `forOputput`?
- One that points to the actual value. `hasValue`?

The text is a bit confused at the moment. How about:
has Input Value - Assigns a value (`hasValue`) to an input (`forInput`) defined by the Procedure that is implemented by an Execution (Actuation, Observation, Sampling).

dr-shorthair:
Thank you for the suggested improvement to the wording. 

I'm reluctant to propose a detailed content model for the object of `hasInputValue` and `hasOutputValue` without implementation experience. These terms will probably need to be tagged `non-normative` anyway. If you do want to push this further, then can you do it in concrete form through a PR?

dr-shorthair:
closes #105

dr-shorthair:
@hylkevds to push this further can you make a PR?

dr-shorthair:
This PR must be merged prior to the major modularization work suggested in https://github.com/w3c/sdw-sosa-ssn/issues/183#issuecomment-1913860023 

We can revisit the hasInputValue issue later if required. (@hylkevds you can create a new issue for that)

hylkevds (approved):

ssn/index.html
```diff
@@ -3581,17 +3611,133 @@ <h5><a href="#SOSAhasOutput">sosa:hasOutput</a></h5>
...
-          <h5><a href="#SOSAProcedureExecution">sosa:ProcedureExecution</a></h5>
-          <p class="crossreference"><strong>IRI:</strong> http://www.w3.org/ns/sosa/ProcedureExecution</p>
+        <section class="informative specterm" id="SOSAExecution" about="sosa:Execution" typeof="owl:Class rdfs:Class">
+          <h5><a href="#SOSAExecution">sosa:Execution</a></h5>
+          <p class="crossreference"><strong>IRI:</strong> http://www.w3.org/ns/sosa/Execution</p>
           <p><strong>an OWL Class</strong></p>
-
+          <p class="note">Execution added in the 2024 Update.<br><br>This term is non-normative pending further
+            implementation experience</p>
           <em property="rdfs:label">Procedure Execution</em> -
```
ldesousa: Shouldn't it be just "Execution"?

ssn/index.html
```diff
@@ -3581,17 +3611,133 @@ <h5><a href="#SOSAhasOutput">sosa:hasOutput</a></h5>
         </section>
 
 
-        <section class="specterm" id="SOSAProcedureExecution" about="sosa:ProcedureExecution"
-          typeof="owl:Class rdfs:Class">
-          <h5><a href="#SOSAProcedureExecution">sosa:ProcedureExecution</a></h5>
-          <p class="crossreference"><strong>IRI:</strong> http://www.w3.org/ns/sosa/ProcedureExecution</p>
+        <section class="informative specterm" id="SOSAExecution" about="sosa:Execution" typeof="owl:Class rdfs:Class">
```
ldesousa, line 3614: This class (and the subsequent properties) are a bit lost in this section (`Procedure`). Since `Observation`, `Actuation` and `Sampling` are sub-classes, `Execution` could justify its own section in the HTML. As of now it would be a fully non-normative section. Could also be expanded with text justifying its introduction.

Thoughts?

ssn/integrated/sosa-oms.ttl
```diff
@@ -272,6 +272,7 @@ sosa-oms:observationType a owl:ObjectProperty ;
     skos:definition "Relation between an Observation and the Platform the Sensor was attached to at the time the observation was made."@en ;
     schema:domainIncludes sosa:Observation ;
     schema:domainIncludes sosa:ObservationCollection ;
+    schema:domainIncludes sosa:ProcedureExecution ;
```
ldesousa: Should be `sosa:Execution`.

ssn/integrated/ssn.ttl
```diff
@@ -356,19 +357,22 @@ sosa:hasUltimateFeatureOfInterest
 sosa:isFeatureOfInterestOf
   rdfs:isDefinedBy sosa: .
 
-sosa:ProcedureExecution a owl:Class ;
-  owl:equivalentClass ssn:ProcedureExecution ;
+sosa:Execution a owl:Class ;
   rdfs:label "Procedure Execution"@en ;
```
ldesousa: Or just "Execution"?

alexrobin:
I personally liked the name `ProcedureExecution` better.
I find the `Execution` term too vague given that we are talking about a very specific type of execution.
I don't think having a longer name really matters since this class is not meant to be used directly.

I agree with @dr-shorthair that we should revisit the definition of input/output at a later stage (separate PR) since I feel it can get complicated...

maximelefrancois86 (approved):

maximelefrancois86:
> `hasInputValue` and `hasOutputValue` will need two properties:
> 
> * One that points to what the `hasInput` / `hasOutput` of the Procedure points to. How about `forInput` / `forOputput`?
> * One that points to the actual value. `hasValue`?
> 
> The text is a bit confused at the moment. How about: has Input Value - Assigns a value (`hasValue`) to an input (`forInput`) defined by the Procedure that is implemented by an Execution (Actuation, Observation, Sampling).

We would solve this problem differently in SAREF ([ETSI TS 103264 V3.2.1](https://www.etsi.org/deliver/etsi_ts/103200_103299/103264/03.02.01_60/ts_103264v030201p.pdf)): applications would define different sub-properties of hasInput to distinguish the different inputs, and these sub-properties would apply to commands and to executions of that command. In the case of controlling an oven, inputs could be: `ex:hasPhase1TemperatureIncreaseRate`, `ex:hasPhase1Duration`, `ex:hasPhase2ConstantTemperature`, `ex:hasPhase2Duration` ...

ssn/integrated/sosa.ttl
```diff
@@ -382,16 +386,37 @@ sosa:isFeatureOfInterestOf a owl:ObjectProperty ;
...
+  schema:domainIncludes sosa:Actuation ;
+  schema:domainIncludes sosa:Execution ;
+  schema:domainIncludes sosa:Observation ;
+  schema:domainIncludes sosa:Sampling ;
+  rdfs:isDefinedBy sosa: .
+
 
-sosa:ProcedureExecution a rdfs:Class , owl:Class ;
+sosa:Execution a rdfs:Class , owl:Class ;
   rdfs:label "Procedure Execution"@en ;
```
ldesousa: "Execution" instead?

ssn/index.html
ldesousa, line 1: ReSpec is flagging one error and four warning when this file is rendered. Anything to worry about?

ldesousa (dismissed):

ssn/index.html
dr-shorthair: The Label does not have to match the class name, else there would be no need for a label. 
However, I've adjusted it to 'Execution of Procedure'

ssn/index.html
dr-shorthair: I agree that the position of sosa:Execution in the document could be changed. 
Let's look at this in the context of the larger reorganisation that is being considered.

ssn/integrated/ssn.ttl
dr-shorthair: See above.

ssn/integrated/sosa.ttl
dr-shorthair: Label not needed in this file.

ssn/index.html
dr-shorthair: I've squashed the error. Will look at the warnings in another thread.

PR #196, opened 2024-02-22
Consumer grade sensor example.

oldskeptic:
This is a first example for the use of a InkBird-IBS-TH2 as commonly available on Amazon. This example comes in two parts: the device definition and a specific device instance that is being used to distribute a beverage.  I've run it into the checker a few times but I likely made a typo or two, some of the "in use" URI's won't resolve for a few days.

I am trying to create a third example of a specific device instance marked up as an [ssn:Deployment](https://www.w3.org/TR/vocab-ssn/#SSNDeployment), but I wanted to get the review process started.


rob-metalinkage (changes requested):
examples are explicitly fragements include in spec now - we should have a separate place for informative examples with use case descriptions

sosa: versions of predicates should be used - to see if we have ssn: usages still required.

sgrellet (changes requested):
+1 on the separation between Sensor Type VS Sensor Instance (same wording as in OMS) thanks to the 2 files content.
Highly important to document that fine line.

I feel we should rather introduce proximateFeatureOfInterest than subtyping the observable property (re: <airRelativeHumidityOfProduct> rdfs:subClassOf equipment:airRelativeHumidity)

To me 'airRelativeHumidity' is the same observable property, be it in the beer pack our outside.
It's the proximateFeatureOfInterest role to precise 'airRelativeHumidity' of what.

OTOH some may say it helps associate the property to the feature.

avillar:
Quick note: instead of using ("minting") full URIs for the sensor instances, blank node id's could be employed (i.e., `<thisBatterySensor>` -> `_:thisBatterySensor`).

oldskeptic:
@avillar > Quick note: instead of using ("minting") full URIs for the sensor instances, blank node id's could be employed (i.e., `<thisBatterySensor>` -> `_:thisBatterySensor`).
I'm using <SerialNumber/Sensor> to communicate instance issue directly; this is an example of an in-use sensor, so I want to avoid blank node and point to example.com instead.

KathiSchleidt:
Then why do we have 2 mentions of thisXXXSensor in line 56 of the instance doc?

sosa:hosts <12345/BatterySensor>, <**thisHumiditySensor**>, <**thisTemperatureSensor**> ;

oldskeptic:
@sgrellet  

> I feel we should rather introduce proximateFeatureOfInstance than subtyping the observable property (re: rdfs:subClassOf equipment:airRelativeHumidity)To me 'airRelativeHumidity' is the same observable property, be it in the beer pack our outside. It's the proximateFeatureOfInstance role to precise 'airRelativeHumidity' of what. OTOH some may say it helps associate the property to the feature.

I don't have a reference for aproximateFeatureOfInstance, but I've gotten myself turned around a few times, so noting things here.

Right now we have:
```
<airTemperatureOfProduct> rdfs:subClassOf equipment:airTemperature ;
    ssn:isPropertyOf <00001/acmePorterSixPack> .
```
The alternative is to use something like
```
equipment:airTemperature ssn:isProxyFor <00001/acmePorterSixPack> .
```
which means the air temperature everywhere determines the product temperature, so we're really trying to say that the air temperature in the six pack is the proxy for the temperature of the six pack, which would be?
```
equipment:airTemperatureInProduct ssn:isProxyFor <00001/acmePorterSixPackTemperature> .    
<00001/acmePorterSixPackTemperature> ssn:isPropertyOf <00001/acmePorterSixPack> .
```
which becomes messy because I'm trying to define the air temperature within the box which is the proxy for the temperature of the box, which is the proxy of the temperature of each of the beer bottles, which is the proxy for the beer within each bottles. While scientifically this is rational way to do it, it's a level of pedantic that I don't see being very popular. 

Willing to change, need a better solution.

oldskeptic:
> Then why do we have 2 mentions of thisXXXSensor in line 56 of the instance doc?
> 
> sosa:hosts <12345/BatterySensor>, <**thisHumiditySensor**>, <**thisTemperatureSensor**> ;

Because the contributor is an unreliable sort. ;) Fixed.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
```diff
@@ -0,0 +1,224 @@
...
+    rdfs:label "IBSTH2 Battery. 2 AAA Batteries. 3.0V Nominal."@en ;
+    skos:editorialNote "Alice doesn't quite believe the battery lifetime claims, so she's tracking every set of batteries she uses and a new battery instances is generated each time." ;
+    gs1:brand <AcmeBatteryBrand> ;
+    dcterms:created "2024-02-20T01:35:45Z"^^xsd:dateTime .
+
+<AcmeBatteryBrand> a gs1:Brand ;
+    rdfs:label "Acme Branded Battery"@en ;
+    gs1:brandName "Acme"@en .
+
+<12345/BatterySensor> a equipment:IBSTH2BatteryVoltageSensor;
```
ldesousa: The battery pack seems more as a FoI than a Property. The Property could be something like "Charge".

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
```diff
@@ -0,0 +1,224 @@
...
+    rdf:comment "This is the instance of the physical sensor instance."@en ;
+    sosa:observes <2024/02/batteryPack> .
+
+<12345/BatterysHealth> rdfs:subClass equipment:IBSTH2BatteryHealth ;
```
ldesousa: Imho this should be the subject of the 12345/BatterySensor Sensor.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
```diff
@@ -0,0 +1,224 @@
...
+
+<airTemperatureOfProduct> rdfs:subClassOf equipment:airTemperature ;
+    ssn:isPropertyOf <00001/acmePorterSixPack> .
+
+<airRelativeHumidityOfProduct> rdfs:subClassOf equipment:airRelativeHumidity;
+    ssn:isPropertyOf <00001/acmePorterSixPack> .
+#
+# Sensor activates and records temperature while in the brewery cooler.
+    
+<observation1a> a sosa:Observation ;
```
ldesousa: This is good. However, I would rather have each of the subjects declared above as instances of SOSA classes.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
```diff
@@ -0,0 +1,224 @@
...
+    prov:atLocation <acmeTruck> ;
+    prov:generated <observation67a> ;
+    prov:generated <observation67b> ;
+    prov:generated <observation67c> ;
+    prov:startedAtTime "2024-02-22T04:15:05Z"^^xsd:dateTime ;
+    prov:endedAtTime "2024-02-22T05:55:38Z"^^xsd:dateTime ;
+    prov:wasStartedBy <driver> ;
+    prov:atLocation <parkingLot> .
+
+<parkingLot> a prov:Location, geosparql:Geometry;
```
ldesousa: A parking lot would rather be an instance of geosparql:Feature (or better one of its Point sub-class), with an associated instance of geosparql:Geometry.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
```diff
@@ -0,0 +1,224 @@
...
+    ssn:isPropertyOf <00001/acmePorterSixPack> .
+
+<airRelativeHumidityOfProduct> rdfs:subClassOf equipment:airRelativeHumidity;
+    ssn:isPropertyOf <00001/acmePorterSixPack> .
+#
+# Sensor activates and records temperature while in the brewery cooler.
+    
+<observation1a> a sosa:Observation ;
+    rel:own <acmeBreweryCo> ;
+    sosa:observedProperty <airTemperatureOfProduct> ;
```
ldesousa: `airTemperatureOfProduct` is a class, not an instance.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
```diff
@@ -0,0 +1,224 @@
...
+    ssn:isPropertyOf <2024/02/batteryPack> .
+
+<12345/HumiditySensor> a equipment:IBSTH2HumiditySensor ;
+    rdf:comment "This is the instance of the humidity sensor instance."@en .
+
+<12345/TemperatureSensor> a equipment:IBSTH2TemperatureSensor ;
+    rdf:comment "This is the instance of the temperature sensor instance."@en .
+
+<airTemperatureOfProduct> rdfs:subClassOf equipment:airTemperature ;
+    ssn:isPropertyOf <00001/acmePorterSixPack> .
```
ldesousa: Although it is not forbidden by the ontology, `isPropertyOf` is meant to have an instance as range, not a class.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
```diff
@@ -0,0 +1,224 @@
...
+    sosa:observedProperty <airRelativeHumidityOfProduct> ;
+    sosa:madeBySensor <12345/HumiditySensor>  ;
+    sosa:hasFeatureOfInterest <00001/acmePorterSixPack> ;
+    sosa:resultTime "2024-02-22T04:15:05Z"^^xsd:dateTime ;
+    sosa:hasSimpleResult "74"^^unit:PERCENT .
+
+<observation67c> rdf:type sosa:Observation,prov:Entity ;
+    rel:own <acmeBreweryCo> ;
+    rel:own <acmeDelivery> ;
+    sosa:observedProperty  <2024/02/batteryPack>;
```
ldesousa: In my view `2024/02/batteryPack` is a FoI, not a property.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
```diff
@@ -0,0 +1,224 @@
...
+    sosa:madeBySensor <12345/BatterySensor>  ;
+    sosa:hasFeatureOfInterest <12345/BatterysHealth> ;
```
ldesousa: This is a class, not instance. Also, an instance of `12345/BatteryHealth` would rather be the subject of `sosa:observedProperty` (with an instance of `2024/02/batteryPack` being the FoI).

ldesousa (changes requested):
@oldskeptic I like this example because the domain/context is straightforwards to understand, beer, temperature, batteries. However, there are some relevant issues to address:
- Properties and FoIs are at times swapped (I noted only some, needs a full review).
- Various subjects of SOSA predicates are classes instead of instances (needs full review too).
- GeoSPARQL is not used in a complete manner.
- Ontology file needs to be identified as such (InkBird-IBS-TH2.ttl).

I also suggest adding a comment referring to the ontology at the top of the graph file.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-deployment.ttl
```diff
@@ -0,0 +1,71 @@
...
+
+@base <http://example.org/data/> .
+
+# This example represents the same data using deployments on platforms, which can make
+# the modeling a bit more intuitive.
+
+
+# Alice packaged the sensor in the box.
+
+<acmePorterSixPackDeployment> a prov:Activity, ssn:Deployment;
```
oldskeptic, line 38: I'd like some comment as to whether this makes sense to everyone.  Ontologically, I can't find a restriction that prevents it through it is a bit counter-intuitive.

Since the sensor and the product share an environment the ability to stack deployment in time and space would allow certain deployments to be used as catch-alls

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
oldskeptic: Resolved in d7d333d

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
oldskeptic: Resolved in d7d333d.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
oldskeptic: Resolved in 29c1ff3.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
oldskeptic: You're reading more ontology semantics in the URL name than I expected to. The average truck GPS just spits out location; in keeping with this I will can the URL to <GPSPosition> to keep it simple.

oldskeptic: Resolved in 09520de.

dr-shorthair:
The correct name for the property is **proximateFeatureOfInterest**
It is in **OMS** - https://docs.ogc.org/as/20-082r4/20-082r4.html#_association_proximatefeatureofinterest . 

In SSN the corresponding predicate is `hasFeatureOfInterest`. 
This is the original predicate that was _always_ intended to point to the _proximate_ FoI - https://w3c.github.io/sdw-sosa-ssn/ssn/#OMS-Alignment-Property . 

In ssn-ext we introduced `hasUltimateFeatureOfInterest` to cover the other usage. 
This has now been added to SSN for the new edition. 
By implication the original predicate covers the 'proximate' case.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
oldskeptic: I'm using 'health' as the sensor documentation has no information as to what it is actually measuring. Likely, it is a linear regression based on the battery voltage, but this is beyond the scope of the problem. FOI fixed.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-deployment.ttl
oldskeptic: As discussed in Feb 29th teleconference, this is ok.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
```diff
@@ -0,0 +1,256 @@
...
+
+#
+# Sensor activates and records temperature while in the brewery cooler.
+    
+<observation1a> a sosa:Observation ;
+    rel:own <acmeBreweryCo> ;
+    sosa:observedProperty equipment:airTemperature ;
+    sosa:hasUltimateFeatureOfInterest <BeerTemperature> ; 
+    sosa:madeBySensor <12345/TemperatureSensor>  ;
+    sosa:hasFeatureOfInterest <00001/acmePorterSixPackAirSample> ;
```
oldskeptic: [ldesousa](https://github.com/ldesousa) Can you please review? This example becomes annoying since the only reference to the beer carton is through the air sample instance.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
oldskeptic: @ldesousa Please review?

dr-shorthair (approved):
Doesn't impact any of the normative artefacts so may as well add to the Git repo

rob-metalinkage (approved):

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
```diff
@@ -171,19 +180,19 @@
     rel:own <acmeBreweryCo> ;
     rel:own <acmeDelivery> ;
     sosa:observedProperty equipment:airTemperature ;
-    sosa:hasUltimateFeatureOfInterest <PO202402/BeerTemperature> ;
+    sosa:hasUltimateFeatureOfInterest <10/PO202402/21/0001/BeerTemperature> ;
```
oldskeptic: @sgrellet Please sanity check my Ultimate Feature of Interest?

hylkevds: I would say the UltimateFeatureOfInterest is `<10/PO202402/21/0001/acmePorterSixPack>`, or even the beer in a specific bottle in the sixpack, since that is what the BeerTemperature is a property of. The BeerTemperature is an ObservedProperty: `<10/PO202402/21/0001/BeerTemperature> a sosa:ObservableProperty` so it can't be a Feature...?

oldskeptic: Fixed with Samples.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
oldskeptic: Changed,

oldskeptic:
@ldesousa @sgrellet Please re-check.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
```diff
@@ -0,0 +1,267 @@
...
+
+<acmePorterSixPack> a beer:Porter, gs1:Product, schema:Product ;
+    skos:definition "A six pack of Acme Porter."@en .
+
+# A specific (instance) six pack of Acme Porter Beers
+# We use GS1-style lot and serial numbers as part of the 
+# identifier to highlight that the carton is a physical instance.
+ 
+<10/PO202402/21/0001/acmePorterSixPack> a beer:Porter, gs1:Product ;
+    rdfs:subClassOf <acmePorterSixPack>;
```
ldesousa, line 62: This should probably be an instance of `<acmePorterSixPack>` instead.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
```diff
@@ -0,0 +1,267 @@
...
+# A six pack as packaged and sold by Acme Brewery Co.
+
+<acmePorterSixPack> a beer:Porter, gs1:Product, schema:Product ;
+    skos:definition "A six pack of Acme Porter."@en .
+
+# A specific (instance) six pack of Acme Porter Beers
+# We use GS1-style lot and serial numbers as part of the 
+# identifier to highlight that the carton is a physical instance.
+ 
+<10/PO202402/21/0001/acmePorterSixPack> a beer:Porter, gs1:Product ;
```
ldesousa, line 61: Would be nicer with the numbers trailing the `/acmePorterSixPack` bit, like the other instances, e.g. `<acmePorterSixPack/10/PO202402/21/0001>`. Ignore if it is intentional.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
```diff
@@ -0,0 +1,267 @@
...
+    prov:generated <observation/67c> ;
+    prov:startedAtTime "2024-02-22T04:15:05Z"^^xsd:dateTime ;
+    prov:endedAtTime "2024-02-22T05:55:38Z"^^xsd:dateTime ;
+    prov:wasStartedBy <driver/1> ;
+    prov:atLocation <gpsLocation/1> .
+
+# This is the output from the truck's GPS receiver. We only get coordinates
+# from it.
+
+<gpsLocation/1> a prov:Location, geosparql:Geometry;
```
ldesousa, line 226: Neater if it is an instance of geosparql:Point instead.

ssn/integrated/informativeExamples/InkBird-IBS-TH2.ttl
```diff
@@ -0,0 +1,134 @@
...
+# representing the commercial sensor product from which different instantions can be derived for each specific physical instance of
+# the product.
+#
+# This prevents duplication of the information required for each new physical instance while allowing results to be related to the
+# actualy physical device that generated them.
+# 
+
+<InkBird-IBS-TH2> a owl:Class, gs1:Product ;
+    rdfs:label "Inkbird IBS-TH2"@en ;
+    rdfs:subClassOf sosa:Platform, equipment:Equipment, <https://w3id.org/seas/BluetoothCommunicationDevice> ;
```
ldesousa, line 27: I am not familiar with gs1, but I think this class should be declared as a sub-class of gs1:Product, not as an instance of gs1:Product.

ssn/integrated/informativeExamples/InkBird-IBS-TH2.ttl
```diff
@@ -0,0 +1,134 @@
...
+<IBSTH2TemperatureSensorLimits> a ssn-system:OperatingRange ;
+    rdfs:label "Inkbird IBS-TH2 Temperature Sensor Limits"@en ;
+    ssn-system:inCondition <IBSTH2TemperatureSensorNormalLimits> .
+
+<IBSTH2TemperatureSensorNormalLimits> a schema:PropertyValue, ssn-system:Condition ;
+    schema:maxValue "60" ;
+    schema:minValue "-40" ;
+    schema:unitCode unit:DEG_C .
+
+<InkBird-IBS-TH2-Procedure> a sosa:Procedure ;
```
ldesousa, line 121: Could this be a SamplingProcedure instead?

ssn/integrated/informativeExamples/InkBird-IBS-TH2.ttl
```diff
@@ -0,0 +1,134 @@
...
+    rdfs:label "Device Hardware Address"@en ;
+    rdfs:range <InkBird-IBS-TH2> .
+
+<IBSTH2Battery> a owl:Class ;
+    rdfs:label "IBSTH2 Battery. 2 AAA Batteries. 3.0V Nominal."@en ;
+    skos:definition "A generic description of the battery pack that needs to be installed in it."@en ;
+    rdfs:subClassOf sosa:FeatureOfInterest .
+
+<IBSTH2BatteryLimits> a schema:PropertyValue, sosa:Property, ssn-system:Accuracy ;
+    skos:editorialNote "Of course, we'd never receive a 0% battery reading..."@en ;
```
ldesousa, line 48: :)

ldesousa (changes requested):
Generally looks good. A few doubts on my side here and there. Let's try getting approved ahead of next meeting.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
oldskeptic: There is no geosparql:Point, you're thinking of a GML Point... There is a geosparql:Centroid which isn't quite what a GPS reports.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
oldskeptic: Intetional. This is a GS1 URL spec.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
ldesousa: Sorry, was referring to `sf:Point` from the [Sampling Features](https://opengeospatial.github.io/ogc-geosparql/geosparql11/sf_geometries.ttl) ontology. It as a sub-class of `geo:Geometry`.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
oldskeptic: "This is the way" according to the [GS1 Digital Link Standard](https://www.gs1.org/docs/Digital-Link/GS1_Digital_link_Standard_i1.1.pdf), page 130. I've tweaked the URLs for this part to conform exactly to the standard.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
oldskeptic: It is done.

ssn/integrated/informativeExamples/InkBird-IBS-TH2.ttl
oldskeptic: Agreed.

ssn/integrated/informativeExamples/InkBird-IBS-TH2.ttl
oldskeptic: I am unsure what is the right thing to do here. The document is about the operating characteristics of the entire platform and reports the precision and accuracy of both sensors. 

Common sense implies that the platform is "sampling" the ambient air, but this document is not about that.

Making a note of this for tomorrow's teleconference.

ssn/integrated/informativeExamples/InkBird-IBS-TH2-instance.ttl
dr-shorthair: Yes, I also prefer the a-box data to be implemented as individuals rather than as classes. 

But I guess we are in part-number vs serial-number territory here, datasheet vs deployed sensor, etc. 
Not sure there is a single answer.

dr-shorthair:
I've been taking a closer look at the InkBird example, also with the [proposed addition of `SystemKind`](https://github.com/w3c/sdw-sosa-ssn/pull/229) in mind. 
I confess I'm a bit baffled by the patterns used. 

Looking at https://w3c.github.io/sdw-sosa-ssn/ssn/#overview-and-examples I would expect the general pattern to look something like this: 

```turtle
<ABCD> a sosa:System , sosa:Sensor , gs1:Product , equipment:Equipment ;
    gs1:pip <https://inkbird.com/products/hygrometer-ibs-th2> ;
    sosa:observes <airRelativeHumidity> ;
    system:hasSystemCapability [ 
        a system:SystemCapability ;
        system:hasSystemProperty [
            a system:MeasurementRange ;
            schema:maxValue "100"^^unit:PERCENT ;
            schema:minValue "0"^^unit:PERCENT ;
        ] ;
        system:hasSystemProperty [
            a system:Resolution ;
            schema:value "2"^^unit:PERCENT ;
        ] ;

    ] ;
    system:hasSurvivalRange [
        a system:SurvivalRange ;
        system:hasSurvivalProperty <EFGH> ;
    ] ;
    system:hasOperatingRange [
        a system:OperatingRange ;
        system:hasOperatingProperty <IJKL> ;
    ] ;
.
```

I've used blank-nodes so that the pattern is clear, but these could be spilled out into named nodes of course. 

It is true that a content model for the classes carrying the actual values is not provided, so for `system:MeasurementRange` I have followed @oldskeptic and used `schema:maxValue` and `schema:minValue` etc
And I like the use of `gs1:pip` to link to an external spec sheet, though I think it could be `dcterms:source` or even `dcterms:isFormatOf`. 

Else this could be formulated as `<ABCD> a system:SystemKind .` in which case this RDF document _is_ the specsheet (though a `dcterms:source` link to an external source is still warranted of course). 

But in the current examples @oldskeptic seems to have weaved a complex web that does not actually leverage the System Capabilities module in the way it was designed. 

Or am I missing something @alexrobin @kjano ?

oldskeptic:
@dr-shorthair I based most of this on the 'DHT22 Description' example. I think we need to document a little better how these patterns are supposed to work. Wrt to #220 and today's teleconference we may need to revisit or properly label some of this. 

> And I like the use of `gs1:pip` to link to an external spec sheet, though I think it could be `dcterms:source` or even `dcterms:isFormatOf`.

gs1:pip is the commercial "glossy" description that attaches to gs1:Product and/or schema:Product. The specsheet is technical in nature and I'd like to attach it to the sensor themselves.

dr-shorthair:
> @dr-shorthair I based most of this on the 'DHT22 Description' example. I think we need to document a little better how these patterns are supposed to work. 

I've had a go at simplifying DHT22 a bit - see https://github.com/w3c/sdw-sosa-ssn/blob/complete-examples/ssn/rdf/examples/dht22.ttl 

Could this be a better template for InkBird-IB-TH2 ?

oldskeptic:
This PR on hold until we resolve a few other issues.

oldskeptic:
This PR is going to be broken up into smaller ones for simplicity.

dr-shorthair:
The Inkbird sensor is used in the Complex System example here: https://w3c.github.io/sdw-sosa-ssn/ssn/#ComplexSystems 

(RDF code at https://github.com/w3c/sdw-sosa-ssn/blob/gh-pages/ssn/rdf/examples/IBS-TH2-PLUS-brief.ttl and https://github.com/w3c/sdw-sosa-ssn/blob/gh-pages/ssn/rdf/examples/IBS-TH2-PLUS.ttl ) 

Documentation is in https://github.com/w3c/sdw-sosa-ssn/blob/gh-pages/ssn/chapters/ModelComplexSystem.html 

What needs to be added?

oldskeptic:
Pull request being closed in favour of a new pull request.

Issue #107, opened 2023-10-25
System types and individuals


maximelefrancois86:
It is still unclear when one wants to model a generic kind of system (aka, model, category, class, ...), whether they should model it as an instance of `ssn:System`, or a subclass of `ssn:System`.

As `ssn:System` is a superclass of `sosa:Sensor`, `sosa:Actuator`, `sosa:Sampler`, this confusion also applies to these class.

This ambiguity is reported in https://www.w3.org/TR/vocab-ssn/#SSNSystem-instances

In the SOSA/SSN examples, systems/sensors/actuators sometimes model generic sensor _kinds_, and sometimes a unique _actual instance_ of sensor:

- B.1 iPhone Barometer  describes `<sensor/35-207306-844818-0/BMP282>` as the _actual instance_ of a BMP282 high-precision barometric pressure sensor hosted on the iphone with IMEI 35-207306-844818-0.
- B.2 Coal Oil Point Reserve describes `<HUMICAP-H>` which is a _kind_ of Vaisala HUMICAP H-chip, a model of accurate temperature and relative humidity probe
- B.12 Examples from O&M refers to a a [graph corresponding to the OMXML examples](https://www.w3.org/TR/vocab-ssn/integrated/examples/om-20.ttl). Four sensors are referred to: `<http://www.csiro.au/people/ps205.html>`, `<http://www.example.org/register/process/scales34.xml>`, `<http://www.example.org/party/individual/abc123>`, `<urn:ogc:object:feature:Sensor:NASA:xyz345>`

It would help a lot to be able to identify _kinds_ of systems generically, the same way we do with property kinds. One advantage would be that we could reuse existing code lists, vocabularies, and taxonomies. Identifiers could be for example:
- Uniclass (ex: [Pr_75_50_76_88  Toxic and combustible gas detectors](https://uniclass.thenbs.com/taxon/pr_75_50_76_88))
- Harmonized Tariff Schedule (HTS) Code (ex: HTS Code: [9025.80.1000 for Electrical Hydrometers And Similar Floating Instruments, Thermometers, Pyrometers, Hygrometers And Psychrometers](https://www.datamyne.com/hts/90/9025801000))
- eClass
- Global Trade Item Numbers (GTINs) , Universal Product Codes (UPCs) , European Article Number (EAN),  etc.
    - see mapping to URNs: https://www.gs1.org/edi-xml/technical-user-guide/5_GS1_namespaces   
[...]

PR #201, opened 2024-03-07
197 common modeling questions

### LINKED ISSUE #197: Common modeling questions
The section 'Common modeling questions' presents some examples of usage which give guidance as to how to apply SSN in applications. 
However, it is problematic in a number of ways:
1. in some cases (e.g. UoM) several options are presented without a strong recommendation as to which one is preferred for interoperability
2. the patterns are just given as examples, and are not formalized as profiles, potentially validatable
3. the set of patterns only covers a small part of the scope of potential variation in SSN usage

Since SSN is quite an abstract ontology, guidance on usage for applications is important. 
However, the scale and scope of the section in the current document is woefully inadequate. 

Propose to revise this section as follows: 
1. add commentary reflecting the notes above
2. OGC to develop a register of SSN Profiles targetted at various applications
3. reword the document to give stronger guidance and recommendations for some specific patterns.

dr-shorthair:
Streamlined Common Modeling Questions section with TTL and JSON-LD examples. 
The patterns for location and unit-of-measure are now more prescriptive. 

Preview here: https://raw.githack.com/w3c/sdw-sosa-ssn/197-common-modeling-questions/ssn/index.html#common-modeling-questions


KathiSchleidt:
Are the `@` in the `hasSimpleResult `attributes correct in ssn/examples/2024ex-uomtype.jsonld?

```
    "hasSimpleResult": {
        "@type": "http://qudt.org/vocab/unit/DEG_C",
        "@value": "19.9"
    }
```

hylkevds:
The thing that bothers me about this example is that the number is encoded as a String. Which means the reader can't be sure the writer used decimal point or comma...
Is there a way around this?

KathiSchleidt:
In SOSA we only have phenomenonTime and resultTime, in OMS we also have validTime. Under [8.4 Forecasts](https://raw.githack.com/w3c/sdw-sosa-ssn/197-common-modeling-questions/ssn/index.html#forecasts) I read "A forecast may be represented as an observation where the value of sosa:phenomenonTime is later in time than the sosa:resultTime."

Do we leave this inconsistency, try to explain it, align it...?

KathiSchleidt:
We need to clarify is sosa:System also gets a sosa:SystemOfInterest (parallel to sosa:Property and sosa:PropertyOfInterest), see #107 before we can finalize section [8.6 Generic or Specific Instances of sosa:System](https://raw.githack.com/w3c/sdw-sosa-ssn/197-common-modeling-questions/ssn/index.html#SOSASystem-instances)

Except for the bits listed here, looks good to go

dr-shorthair:
> In SOSA we only have phenomenonTime and resultTime, in OMS we also have validTime. Under [8.4 Forecasts](https://raw.githack.com/w3c/sdw-sosa-ssn/197-common-modeling-questions/ssn/index.html#forecasts) I read "A forecast may be represented as an observation where the value of sosa:phenomenonTime is later in time than the sosa:resultTime."
> 
> Do we leave this inconsistency, try to explain it, align it...?

I don't see the inconsistency. `validTime` might be a useful addition, but the explanation of times for a forecast is correct with just "`sosa:phenomenonTime` is later than `sosa:resultTime`".

dr-shorthair:
> The thing that bothers me about this example is that the number is encoded as a String. Which means the reader can't be sure the writer used decimal point or comma... Is there a way around this?

Yes, I believe that in RDF any literal that is not typed as a number is a string. 
If `http://qudt.org/vocab/unit/DEG_C` was asserted to be a sub-class of `xsd:decimal` then perhaps it would be OK? 

In the JSON encoding if the quotes are dropped from "19.9" does that make it a number?

dr-shorthair:
> We need to clarify is sosa:System also gets a sosa:SystemOfInterest (parallel to sosa:Property and sosa:PropertyOfInterest), see #107 before we can finalize section [8.6 Generic or Specific Instances of sosa:System](https://raw.githack.com/w3c/sdw-sosa-ssn/197-common-modeling-questions/ssn/index.html#SOSASystem-instances)

The conversation seems to be converging. See #209

hylkevds:
> In the JSON encoding if the quotes are dropped from "19.9" does that make it a number?

Yes, without the quotes it's a proper number.

oldskeptic (approved):

rob-metalinkage (approved):

PR #226, opened 2024-05-07
225 235 Add madeBySystem and madeExecution 

### LINKED ISSUE #225: `madeExecution` `madeBySystem` super-properties used on `Execution`
There is an asymmetry between 
- `usedProcedure`, that links any instance of an `Execution` or sub-class to its `Procedure`, 

and 
- `madeByActuator` `madeBySensor` `madeBySampler` that links an instance of a specific `Execution` sub-class  (i.e. `Actuation` `Observation` `Sampling`) to an instance of the `System` subclass (i.e. `Actuator` `Sensor` `Sampler`) that made it. 

Should this be neatened up by introducing `madeBySystem` as the super-property? 
And `usedActuatingProcedure` `usedObservingProcedure` `usedSamplingProcedure` as sub-properties?

dr-shorthair:
to sosa-common, and sub-property axioms to ssn-* modules

Add usedActuatingProcedure usedObservingProcedure usedSamplingProcedure to sosa-* modules, and  sub-property axioms to ssn-* modules 

Add actsOn to sosa-actuation

Closes #225


dr-shorthair:
Terms in red are the proposed additions, classes in brown are changes we already introduced

System-Execution-Procedure-specializations

ldesousa:
I don't see technical issues with this implementation (yet), but I would like to think a bit more on this from higher level of abstraction. The ontology is moving towards a set of four very similar graphs, centered on the four main classes:
- Execution
- Actuation
- Observation
- Sampling

Those classes are well defined and understood, not only by those of us in the group, but also to users in general that may deal with concerned datasets in this context. However, the semantic nuances between the sub-classes of Procedure and Property are less evident, i.e.:
- ActuatingProcedure
- ObservingProcedure
- SamplingProcedure
- ActuableProperty
- ObservableProperty
- ~~SamplingProperty~~

Are all those classes necessary? Do they all have a well defined and unique semantic definition?

Finally coming to the predicates: are they all semantically unique? Or is a single predicate object property between Execution and Procedure enough to convey that relation in all sub-graphs? I would rather have a leaner ontology, with as many elements as possible at the Execution (abstract?) level, only to be specialised in case of objective semantic differences. 

Thoughts?

dr-shorthair:
Thanks @ldesousa - this is exactly the conversation we need. Please continue over in #225

oldskeptic:
> Now we have introduced a generalization `Execution`, and specializations `ActuatingProcedure`, `ObservingProcedure`, `SamplingProcedure` (note that `ActuatableProperty` and `ObservableProperty` were there before). The proposed predicates (in red above) are really just my chasing through the consequences.

Do we need the specializations `ActuatingProcedure`, `ObservingProcedure`, `SamplingProcedure`? Is there a material difference in these classes that wouldn't be covered by `Procedure`? A bit late for this question I realise.

dr-shorthair:
I've copied this discussion back to issue #225 - better to have it there rather than on the PR. 
Please lets keep the high-level discussion there. 
PR comments should only be on whether the PR addresses the agreed resolution.

dr-shorthair:
Preview https://raw.githack.com/w3c/sdw-sosa-ssn/225-madebysystem/ssn/index.html

Closes #225 
Closes #235

rob-metalinkage (dismissed):
I think the text clarification should be emphasised as SHOULD NOT

dr-shorthair:
@rob-metalinkage - fixed

ssn/rdf/ontology/core/sosa-common.ttl
```diff
@@ -229,7 +229,14 @@ sosa:Execution
...
   skos:definition """
-  Act of carrying out a Procedure. This class generalizes the Observation, Actuation, and Sampling classes. It is not expected to be used directly.
+  Act of carrying out a Procedure. 
+  """@en ;
+  skos:scopeNote """
+  Execution generalizes the Actuation, Observation, and Sampling classes. 
+  It appears in the ontology to support the description of a common pattern for executions and their properties. 
+
+  Execution may be considered to be abstract, so individuals SHOULD NOT be declared to be members of only this class. 
+  Applications SHOULD use members of concrete sub-classes of Execution.
```
ldesousa: The term "member" is not that common in the Semantic Web. I suggest being more precise here, using instead "object properties". Besides, the most important sentence is the second, this third sentence could even be absent.

ssn/rdf/ontology/core/sosa-common.ttl
```diff
@@ -341,6 +348,24 @@ sosa:isResultOf
...
+  a owl:ObjectProperty ;
+  rdfs:label "made by system"@en ;
+  skos:definition """
+  Relation linking an Execution to the System that made it.
+  """@en ;
+  skos:scopeNote """
+  madeBySystem generalizes the madeByActuator, madeBySensor, and madeBySampler properties. 
+  It appears in the ontology to support the description of a common pattern for executions and their properties. 
+
+  madeBySystem may be considered to be abstract, so SHOULD NOT appear in data instances.
```
ldesousa: "may be" and "SHOULD NOT" possibly not aligned. I would rather have "should be considered" or even "must be considered".

ssn/rdf/ontology/core/sosa-common.ttl
```diff
@@ -371,11 +396,11 @@ sosa:phenomenonTime
   skos:scopeNote """
   The phenomenonTime is not necessarily the same as the resultTime. 
   
-  In the case of Actuation, phenomenonTime should indicate the time that the actuator was active. 
+  In the case of Actuation, phenomenonTime SHOULD indicate the time that the actuator was active.
```
ldesousa: "the time at which the actuator was active"

ssn/rdf/ontology/core/sosa-common.ttl
```diff
@@ -371,11 +396,11 @@ sosa:phenomenonTime
   skos:scopeNote """
   The phenomenonTime is not necessarily the same as the resultTime. 
   
-  In the case of Actuation, phenomenonTime should indicate the time that the actuator was active. 
+  In the case of Actuation, phenomenonTime SHOULD indicate the time that the actuator was active. 
 
   In the case of Observation, the phenomenonTime may be distant from the resultTime, for example in the case of forecasts the phenomenonTime is after the resultTime, and in historical, geological and cosmological investigations the phenomenonTime may be far in the past while the resultTime is contemporary. 
 
-  In the case of Sampling the phenomenonTime should indicate the period during which the Sampling Procedure was active, while the resultTime may indicate when the sample comes into the posession of the operator. 
+  In the case of Sampling the phenomenonTime SHOULD indicate the period during which the Sampling Procedure was active, while the resultTime may indicate when the sample comes into the posession of the operator.
```
ldesousa: This sentence is too long, imho. I suggest this alternative formulation:
"In the case of Sampling, phenomenonTime SHOULD indicate the period during which the Sampling Procedure was active. Whereas resultTime may indicate when the sample comes into the possession of the operator."

ssn/rdf/ontology/core/sosa-common.ttl
```diff
@@ -443,8 +468,16 @@ sosa:System
...
   skos:definition """
-  System is a unit of abstraction for pieces of infrastructure that implement Procedures. A System may have components, its subsystems, which are other systems.
+  System is a piece of infrastructure that implements a Procedure. 
+  A System may have components, its subsystems, which are other systems.
   """@en ; 
+  skos:scopeNote """
+  System generalizes the Actuator, Sensor, and Sampler classes. 
+  It appears in the ontology to support the description of a common pattern for executions and their properties. 
+
+  System may be considered to be abstract, so individuals SHOULD NOT be declared to be members of only this class.
```
ldesousa: "instances" instead of "members". That is more common in the Semantic Web.

ssn/rdf/ontology/core/sosa-common.ttl
```diff
@@ -561,6 +594,24 @@ sosa:isSubSystemOf
...
+  a owl:ObjectProperty ;
+  rdfs:label "made execution"@en ;
+  skos:definition """
+  Relation between an System and an Execution it has made.
+  """@en ;
+  skos:scopeNote """
+  madeExecution generalizes the madeActuation, madeObservation, and madeSampling properties. 
+  It appears in the ontology to support the description of a common pattern for executions and their properties. 
+
+  madeExecution may be considered to be abstract, so SHOULD NOT appear in data instances.
```
ldesousa: As before, I suggest replacing "may be" for "should be" or "must be". The architect(s) has a clear intention here.

ssn/rdf/ontology/core/sosa.ttl
```diff
@@ -31,6 +31,24 @@ sosa:
   dcterms:modified "2024-02-03"^^xsd:date ;
   owl:imports sosa-act: , sosa-obs: , sosa-sam:.
 
+sosa:madeBySystem 
+  schema:domainIncludes sosa:Actuation ;
```
ldesousa, line 35: In this cases it is more common to see the use of a list. E.g.:
`schema:domainIncludes [sosa:Actuation, sosa:Observation, sosa:Sampling] ;`
Makes it easier to read, in my view.

ssn/rdf/ontology/core/sosa.ttl
```diff
@@ -31,6 +31,24 @@ sosa:
...
+  schema:domainIncludes sosa:Observation ;
+  schema:domainIncludes sosa:Sampling ;
+  schema:rangeIncludes sosa:Actuator ;
+  schema:rangeIncludes sosa:Sensor ;
+  schema:rangeIncludes sosa:Sampler ;
+  rdfs:isDefinedBy sosa-common: .
+
+sosa:madeExecution 
+  schema:domainIncludes sosa:Actuator ;
```
ldesousa, line 44: Ditto.

ssn/images/Execution.svg
ldesousa, line 1: Would it be simple to convey here the different namespaces (SOSA, SOSA-OMS)? Say with graphical UML elements like packages.

ssn/chapters/Common.html
```diff
@@ -1186,20 +1145,27 @@ <h6><a href="#SOSAExecution">sosa:Execution</a></h6>
...
-                            <span rel="rdfs:isDefinedBy"><a
-                                    href="http://www.w3.org/ns/sosa/common/">http://www.w3.org/ns/sosa/common/</a></span>
+                            <span rel="skos:scopeNote">
+                                Execution generalizes the <a href="#SOSAActuation">Actuation</a>, <a
+                                    href="#SOSAObservation">Observation</a>, and
+                                <a href="#SOSASampling">Sampling</a> classes.
+                                It appears in the ontology to support the description of a common pattern for executions
+                                and their properties.
+
+                                Execution may be considered to be abstract, so individuals SHOULD NOT be declared to be
```
ldesousa: Consider replacing "may" with "should" or "must". See comments on the TTL.

ssn/chapters/Common.html
```diff
@@ -1186,20 +1145,27 @@ <h6><a href="#SOSAExecution">sosa:Execution</a></h6>
...
+                                members of only this class.
+                                Applications SHOULD use members of concrete sub-classes of Execution.
```
ldesousa: "instances" instead of "members". Consistent with comments on the TTL.

ssn/chapters/Common.html
```diff
@@ -1769,6 +1735,93 @@ <h6><a href="#SOSAisResultOf">sosa:isResultOf</a></h6>
...
+                        <th>Scope note</th>
+                        <td>
+                            <code>madeBySystem</code> generalizes the
+                            <a href="#SOSAmadeByActuator">sosa:madeByActuator</a>,
+                            <a href="#SOSAmadeBySensor">madeBySensor</a>, and
+                            <a href="#SOSAmadeBySampler">sosa:madeBySampler</a> properties.
+                            It appears in the ontology to support the description of a common pattern for executions and
+                            their properties.
+
+                            <code>madeBySystem</code> may be considered to be abstract, so SHOULD NOT appear in data
```
ldesousa: "should" or "must" in place of "may".

ssn/chapters/Common.html
```diff
@@ -1903,7 +1956,7 @@ <h5><a href="#SOSAphenomenonTime">sosa:phenomenonTime</a></h5>
                 <br>
                 <br>
                 In the case of <a href="#SOSAActuation">sosa:Actuation</a>, <a
-                    href="#SOSAphenomenonTime">sosa:phenomenonTime</a> should indicate the time that the <a
+                    href="#SOSAphenomenonTime">sosa:phenomenonTime</a> SHOULD indicate the time that the <a
```
ldesousa: "the time at which the sosa:Actuator was active"

ssn/chapters/Common.html
```diff
@@ -2389,11 +2437,27 @@ <h6><a href="#SOSASystem">sosa:System</a></h6>
...
+                        <th>Scope note</th>
+                        <td>
+                            <span rel="skos:scopeNote">
+                                System generalizes the <a href="#SOSAActuator">Actuator</a>, <a
+                                    href="#SOSASensor">Sensor</a>, and
+                                <a href="#SOSASampler">Sampler</a> classes.
+                                It appears in the ontology to support the description of a common pattern for executions
+                                and their properties.
+
+                                System may be considered to be abstract, so individuals SHOULD NOT be declared to be
```
ldesousa: "must" or "should" in place of "may"

ssn/chapters/Common.html
```diff
@@ -2973,6 +3037,86 @@ <h6><a href="#SOSAisSubSystemOf">sosa:isSubSystemOf</a></h6>
...
+                        <th>Scope note</th>
+                        <td>
+                            <code>madeExecution</code> generalizes the
+                            <a href="#SOSAmadeActuation">sosa:madeActuation</a>,
+                            <a href="#SOSAmadeActuation">madeObservation</a>, and
+                            <a href="#SOSAmadeSampling">sosa:madeSampling</a> properties.
+                            It appears in the ontology to support the description of a common pattern for executions and
+                            their properties.
+
+                            <code>madeExecution</code> may be considered to be abstract, so SHOULD NOT appear in data
```
ldesousa: "must" or "should" in place of "may"

ssn/images/System.svg
ldesousa, line 1: There is a problem with this file. Both Gimp and Inkscape complain it has errors. GitHub can't render it.

ldesousa (changes requested):

ssn/images/System.svg
oldskeptic: I'm still having trouble viewing this file again.

ssn/images/Execution.svg
KathiSchleidt, line 1: Still has specialized Procedure Types

ssn/rdf/ontology/core/sosa-common.ttl
dr-shorthair: I don't understand where 'object properties' would appear in the text. 

I used 'members' in the sense of the 'class extension'. 
Every individual of type `ex:Class1` is a _member_ of that class. 

I agree that the two SHOULD sentences are redundant. 
Perhaps it would be better phrased as 
> Execution MUST be considered to be abstract, so any individual Execution SHOULD be declared to be an instance of one of the concrete sub-classes, and not of the Execution class.

ssn/rdf/ontology/core/sosa-common.ttl
dr-shorthair: > madeBySystem MUST be considered to be abstract, so SHOULD NOT appear in data instances.
> Applications can use concrete sub-properties of madeBySystem.

ssn/rdf/ontology/core/sosa-common.ttl
dr-shorthair: OK

ssn/rdf/ontology/core/sosa-common.ttl
dr-shorthair: Sentences broken up:
> In the case of Observation, the phenomenonTime MAY be distant from the resultTime. 
>  For example in the case of forecasts the phenomenonTime is after the resultTime. 
>  In historical, geological and cosmological investigations the phenomenonTime can be far in the past, while the resultTime is contemporary. 
>
>  In the case of Sampling the phenomenonTime SHOULD indicate the time during which the Sampling Procedure was active.
>  The resultTime can indicate when the sample comes into the posession of the operator.

ssn/rdf/ontology/core/sosa-common.ttl
dr-shorthair: Re-worded consistent with `Execution`

>  System MUST be considered to be abstract, so any individual System SHOULD be declared to be an instance of one of the concrete sub-classes, and not of the System class.

ssn/rdf/ontology/core/sosa-common.ttl
dr-shorthair: >   madeExecution MUST be considered to be abstract, so SHOULD NOT appear in data instances. 
>  Applications can use concrete sub-properties of madeExecution.

ssn/rdf/ontology/core/sosa.ttl
dr-shorthair: No `[  ]` required. In TTL a set of objects for the same predicate is simply comma-separated. 

However, readability is not the key concern in the RDF/TTL. 
They are broken onto separate lines for ease of editing/bookkeeping at this stage. 
The RDF/TTL can be re-styled later if desired.

ssn/rdf/ontology/core/sosa.ttl
dr-shorthair: Ditto.

ssn/images/Execution.svg
dr-shorthair: Ignore the SVG - they are not up-to-date. 
I'll remove them (for now).

ssn/chapters/Common.html
dr-shorthair: I've made the HTML consistent with RDF/TTL everywhere.

ssn/chapters/Common.html
dr-shorthair: Yes.

ssn/chapters/Common.html
dr-shorthair: Yes

ssn/chapters/Common.html
dr-shorthair: Yes

ssn/chapters/Common.html
dr-shorthair: Yes

ssn/chapters/Common.html
dr-shorthair: Yes

ssn/images/System.svg
dr-shorthair: Ignore SVG versions. Removed now.

ssn/images/Execution.svg
dr-shorthair: SVG all obsolete. Ignore (removed).

ldesousa (approved):

oldskeptic (approved):

PR #229, opened 2024-05-27
107 move systemkind to system capabilities

### LINKED ISSUE #107: System types and individuals

dr-shorthair:
I've taken Alex's proposal and moved the classes, properties and related axioms into the System Capabilities extension module. 

Preview: https://raw.githack.com/w3c/sdw-sosa-ssn/107-move-systemkind-to-system-capabilities/ssn/index.html#System-capabilities


alexrobin:
I'm fine with this but could we just use the `ssn` namespace (http://www.w3.org/ns/ssn/) for the capabilities module (instead of the old `ssn-system`,  http://www.w3.org/ns/ssn/systems) now that we moved pretty much everything else to `sosa`?

And deprecate concepts in `ssn-system` just like we did for the `ssn` concepts we moved to `sosa`?

dr-shorthair:
@alexrobin interesting proposal. However, I'm not sure it works - 

We 'emptied out' the `ssn:` ns - the terms that were in there are now tagged 'deprecated'.   
The SSN module now only houses the axiomatization of the SOSA modules - as in the previous version. 

The plan is that dereferencing `http://www.w3.org/ns/ssn/` should get an ontology that supports everything that the previous edition did (the deprecated terms are still quietly imported), though it will also import additional terms in the `sosa:` ns. 

So I think it would be confusing to make `ssn:` do double duty, also housing the System Capabilities terms.

dr-shorthair:
Closes #107

dr-shorthair:
Update team agreed not to introduce a specific SystemKind class.

PR #236, opened 2024-07-18
232 domain models vs feature of interest

### LINKED ISSUE #232: Feature-of-Interest or just Feature
Do we want to leave Feature-of-Interest as a class, or reduce this to Feature.

Under STA, we've done exactly this, as we realized that of interest is not an characteristic of the feature itself, it's the role taken in the context of the Observation applied to this Feature.

However, in the semantic context, multiple derivation is far more common. Thus, we could also leave Feature-of-Interest as it is, provide guidance that in implementation, dedicated domain Feature-of-Interest types should be created through derivation from both Feature-of-Interest and the relevant domain feature.

Question: does this mean that one can only apply Observations to a Feature if it's also derived from Feature-of-Interest, or can we utilize any domain Feature here?

dr-shorthair:
Preview https://raw.githack.com/w3c/sdw-sosa-ssn/232-domain-models-vs-feature-of-interest/ssn/index.html#model-domain 

Fixes #232


dr-shorthair:
@KathiSchleidt @sgrellet @hylkevds @rob-metalinkage @avillar @oldskeptic could you review please?

hylkevds (approved):
Looks good.

rob-metalinkage (approved):
These look good - however I still feel there is more room to address the abstraction level of observedProperty - whether and when it is explicitly a property of a domain model.  This should be addressed in a separate refinement however.

oldskeptic (approved):

PR #249, opened 2024-10-01
Deprecate ObservableProperty and ActuatableProperty

dr-shorthair:
Remove ObservableProperty and ActuatableProperty from RDF (TTL) files, from documentation (HTML), and from diagrams. 
Add to ssn-deprecated.ttl. 

Preview: https://raw.githack.com/w3c/sdw-sosa-ssn/237-deprecate-observableproperty-actuatableproperty/ssn/index.html 

Replaces #238 which @oldskeptic created in a fork, which has now drifted too far from the head to be immediately usable. 

Fixes #237

Issue #237, opened 2024-07-30
Deprecate ObservableProperty, ActuatableProperty


dr-shorthair:
Only `sosa:Property` is needed. 
No functional difference between the sub-classes. 
Legacy sub-class relationships in ssn-deprecated.ttl would preserve backward compatibility. 

See argument here: https://github.com/w3c/sdw-sosa-ssn/issues/235#issuecomment-2243218266

dr-shorthair:
@alexrobin @KathiSchleidt @ldesousa any comment on this?

ldesousa:
I am in favour of this change.

dr-shorthair:
Kathi also agreed earlier
2024-09-25 telecon agreed with this.

sgrellet:
fine with me as well

alexrobin:
I support this as well

dr-shorthair:
@alexrobin @sgrellet could you review https://github.com/w3c/sdw-sosa-ssn/pull/249 please?

alexrobin:
The only thing I noticed is this part of the documentation of `Actuator` that I don't think is necessary anymore:
```
Restriction: sosa:forProperty ONLY sosa:Property
```

Same thing in the RDF:
```
sosa:Actuator 
  rdfs:subClassOf sosa:System ;
  rdfs:subClassOf [ a owl:Restriction ; owl:onProperty sosa:forProperty ; owl:allValuesFrom sosa:Property ] ;
```

However, should the `Actuator` class have a restriction on the new `actsOn` property though?

Otherwise, it looks good to me.

dr-shorthair:
Right. That restriction is 'inherited' from the parent class `sosa:System`. 
Also missed similar on `sosa:Sensor`. 
Fixed now. 

@alexrobin @sgrellet could you approve review so it can be merged.

PR #250, opened 2024-10-02
Removes references to ObservableProperty and ActuatableProperty

ldesousa:
Part of the fix to #237. Compounds on PR #249.


dr-shorthair:
I fixed a small number of mentions of OP and AP in the branch for #249. 
Many of the other changes you are proposing are in SSN-Extensions rather than SSN. 
Not needed ... until we carry some of this stuff over into SSN.

ldesousa:
@dr-shorthair Do we still need to look at this? Or can I simply close it?

dr-shorthair:
I think we can close it.

PR #254, opened 2024-10-15
107 systemtype patterns

### LINKED ISSUE #107: System types and individuals

dr-shorthair:
Patterns for System type descriptions, including complex sensors

Fixes #107


oldskeptic (approved):

dr-shorthair:
Previews:
- https://raw.githack.com/w3c/sdw-sosa-ssn/107-systemtype-patterns/ssn/index.html#ComplexSystems 
- https://raw.githack.com/w3c/sdw-sosa-ssn/107-systemtype-patterns/ssn/index.html#SystemType 

@ldesousa @sgrellet could you review?

ssn/chapters/ModelPatterns.html
```diff
@@ -0,0 +1,100 @@
...
+        These are discussed further in the Spatial Data on the Web Best Practices note [[SDW-BP]]. </p>
+    <p>In particular, GeoSPARQL [[GeoSPARQL]] provides a flexible and relatively complete platform for geospatial
+        objects, that fosters interoperability between geo-datasets. To do so, these entities can be declared as
+        instances of geo:Feature and geometries can be assigned to them via the geo:hasGeometry property.
+        In case of classes, e.g., specific features of interests such as rivers, these can be defined as subclasses
+        of
+        geo:Feature.</p>
+    <p>For example, if observations are made of the atmosphere at a specific location, it might be described as a
+        Sample using the following pattern:</p>
+    <pre class="example" data-include="./rdf/examples/2023ex1.ttl" data-include-format="text"></pre>
```
ldesousa, line 27: This example breaks the pattern described above, `earthAtmosphere_StE` is not an instance of `geo:Feature`. Moreover, in the context of *Sampling*, the spatial features would rather be the `Sampler` instance and the physical edifice storing the `Sample` instance.

We could consider a `Sample` instance to have *movement*. But that is a whole other kettle of fish, away from which we can remain.

ssn/chapters/ModelPatterns.html
```diff
@@ -0,0 +1,100 @@
...
+
+<section id="forecasts" class="informative">
+    <h3>Forecasts</h3>
+    <p>One MAY also represent forecasts as observations if the value of <code>sosa:phenomenonTime</code> is later in
+        time than the <code>sosa:resultTime</code>.
+        Given the definition of these terms, it means that: <em>The time when the Observation act was completed is
+            before the time that the Result of the observation applies to the FeatureOfInterest.</em></p>
+    <p class="note"> Other means to represent forecasts are reported, but not in the scope of this specification.
+        For example [[Lefrancois-et-al-2017]] derives the SSN <code>Sensing/Sensor/Observation</code>
+        pattern and define <code>Forecasting/Forecaster/Forecast</code> classes.</p>
```
ldesousa: Second person mixed with third person.

ssn/chapters/ModelPatterns.html
```diff
@@ -0,0 +1,100 @@
...
+        of
+        geo:Feature.</p>
+    <p>For example, if observations are made of the atmosphere at a specific location, it might be described as a
+        Sample using the following pattern:</p>
+    <pre class="example" data-include="./rdf/examples/2023ex1.ttl" data-include-format="text"></pre>
+
+</section>
+
+<section id="forecasts" class="informative">
+    <h3>Forecasts</h3>
```
ldesousa, line 32: This recommendation is logical. However, it took me a while to understand. Consider reviewing/simplifying the text.

ssn/rdf/examples/Thermometer.ttl
```diff
@@ -0,0 +1,54 @@
...
+  sosa:madeBySensor ex:Mums-clinical-thermometer ;
+  sosa:observedProperty qk:Temperature ;
+.
+sensor:Mercury-in-glass-thermometer
+  a owl:Class ;
+  rdfs:comment "Temperature sensor based on observing the the expansion of a column of mercury inside a glass tube" ;
+  rdfs:label "A specific type of temperature sensor" ;
+  rdfs:subClassOf sensor:TemperatureSensor ;
+.
+sensor:Special-Mercury-in-glass-thermometer
```
ldesousa: What is the purpose of this class? Looks like nothing is using it.

ssn/rdf/examples/Thermometer.ttl
```diff
@@ -0,0 +1,54 @@
...
+  a owl:Class ;
+  rdfs:subClassOf sensor:Mercury-in-glass-thermometer ;
+.
+sensor:TemperatureSensor
+  a owl:Class ;
+  rdfs:label "A generic temperature sensor" ;
+  rdfs:subClassOf sosa:Sensor ;
+  rdfs:subClassOf [
```
ldesousa, line 45: This doesn't look right. Instead of owl:hasValue, it should probably be using owl:allValuesFrom.

You can open a new issue an assign it to me. I will look into in more detail.

ssn/rdf/examples/Thermometer.ttl
ldesousa: What is the purpose of this class? Since it is not being used, I suggest removing.

ssn/images/Inkbird-examples-sensorKind.png
ldesousa, line 1: The names `SensorKind` and `SensorInstance` are not the best, as discussed. Is this diagram necessary? If yes I would rather have `InBirdIBS-TH2` as a direct sub-class of `Sensor` and remove the other two.

ssn/images/Inkbird-sensorkind.png
ldesousa, line 1: Same comment on `SensorKind` and `SensorInstance`.

ssn/rdf/examples/InkBird.ttl
ldesousa, line 1: Missing the URI abbreviation section.

ssn/rdf/examples/InkBird.ttl
```diff
@@ -0,0 +1,50 @@
+
+
+<https://example.org/sensor/IBS-TH2> rdfs:subClassOf sosa:Sensor ;
+    rdfs:comment "The class of IBS-TH2 sensors is a subclass of the general class of sensors" ;
+    gs1:pip <https://inkbird.com/products/hygrometer-ibs-th2> ;
+    rdfs:label "Bluetooth Temperature and Humidity Sensor IBS-TH2" ;
+    skos:notation "IBS-TH2" ;
+    sosa:observes <http://qudt.org/vocab/quantitykind/RelativeHumidity> ;
```
ldesousa: The range of sosa:observes is sosa:Sensor, not owl:Class.

ssn/rdf/examples/InkBird.ttl
```diff
@@ -0,0 +1,50 @@
...
+    sosa:observes <http://qudt.org/vocab/quantitykind/Temperature> ;
+    system:hasSystemCapability [
```
ldesousa, line 17: I don't  know `system` that well, but I doubt these object properties have owl:Class as range. Please verify.

ssn/rdf/examples/InkBirdSpecSheet.pdf
ldesousa, line 1: Pay attention to Copyright when adding files such as this one to the repository.

ssn/rdf/ontology/core/sosa-common.ttl
```diff
@@ -500,15 +500,22 @@ sosa:System
...
   """@en ; 
   skos:scopeNote """
   System generalizes the Actuator, Sensor, and Sampler classes. 
   It appears in the ontology to support the description of a common pattern for executions and their properties. 
 
   System MUST be considered to be abstract, so any individual System SHOULD be declared to be an instance of one of the concrete sub-classes, and not of the System class. 
   """@en ;
+  skos:scopeNote """
+  Specific system types may be implemented as sub-classes of this class or its sub-classes.
+  Their descriptions may capture the properties and characteristics of a whole class of systems, corresponding with the information typically found in a data-sheet.
```
ldesousa, line 515: A small example in a single sentence would be helpful here.

ssn/rdf/ontology/core/sosa-common.ttl
```diff
@@ -500,15 +500,22 @@ sosa:System
...
+
+  Individual system instances are members of this class or its sub-classes.
```
ldesousa, line 517: Consider adding here a link to the examples or the Patterns section.

ldesousa (changes requested):

ssn/chapters/ModelPatterns.html
dr-shorthair: 1. The `geo:hasGeometry` property has `rdfs:domain geo:Feature` so this entails that `earthAtmosphere_StE` is of `rdf:type geo:Feature`.  It is not necessary to state it explicitly. And in this case I felt less is more. 

2. Yes, the location could be of the `Sampler`, particularly for in-situ observations. But this example was making a separate point: that a `Sample` may have a location using this simple pattern. This allows the location to be captured directly without having to traverse a Sampling/Sampler path.

ssn/chapters/ModelPatterns.html
dr-shorthair: Sorry - can't see the problem. Can you be more explicit?

ssn/chapters/ModelPatterns.html
dr-shorthair: How about:
> Forecasts MAY be represented as observations if the value of sosa:phenomenonTime is later in time than the sosa:resultTime. This means that the time when the Observation result was obtained is before the time that the result applies to the  FeatureOfInterest.

ssn/rdf/examples/InkBird.ttl
dr-shorthair: The range of `sosa:observes` is `sosa:Property` - https://w3c.github.io/sdw-sosa-ssn/ssn/#SOSAobserves

ssn/rdf/examples/InkBird.ttl
dr-shorthair: The range of `sosa:observes` is `sosa:Property`. 
http://qudt.org/vocab/quantitykind/Temperature is an individual of `rdf:type qudt:QuantityKind`, so from this statement it is inferred to also be `rdf:type sosa:Property` which is fine.

ssn/rdf/examples/InkBirdSpecSheet.pdf
dr-shorthair: I've removed it from the repo.

ssn/rdf/ontology/core/sosa-common.ttl
dr-shorthair: The HTML documentation has a link to the pattern which has examples/. 
See https://raw.githack.com/w3c/sdw-sosa-ssn/107-systemtype-patterns/ssn/index.html#SOSASystem 

If this is not enough please make a concrete proposal.

ssn/rdf/ontology/core/sosa-common.ttl
dr-shorthair: The HTML documentation has a link to the pattern which has examples/.
See https://raw.githack.com/w3c/sdw-sosa-ssn/107-systemtype-patterns/ssn/index.html#SOSASystem

ssn/chapters/ModelPatterns.html
ldesousa: 1. As it stands, that example graph would produce errors when fed to a reasoner. The specification of the [`geo:hasGeometry` property](https://docs.ogc.org/is/22-047r1/22-047r1.html#_01414b7a-8464-4a9b-897d-46c471274ea6) is strict, its domain and ranges are not meant to be implied.

Broader picture. If GeoSPARQL is to fall within the SDWWG's purview, we better make sure to use it as it is meant to be used. IMHO.

ssn/chapters/ModelPatterns.html
ldesousa: "derives" and later "define".

ssn/rdf/examples/InkBird.ttl
ldesousa: Alright. But [quantitykind:RelativeHumidity](http://qudt.org/vocab/quantitykind/RelativeHumidity) is neither.

ssn/rdf/examples/InkBird.ttl
ldesousa: In my understanding of OWL this is not fine. And a reasoner should fail to make that inference.

Perhaps it is best to verify this with a third person that knows OWL better.

ssn/rdf/ontology/core/sosa-common.ttl
ldesousa: The patterns section seems enough.

ldesousa:
@dr-shorthair I would like to clarify the question on `rdfs:domain` (and `rdfs:range`) before giving the green light. OWL defines those as constraints, and [makes it clear](https://www.w3.org/TR/owl-ref/#domain-def):

> An `rdfs:domain` axiom asserts that the subjects of such property statements **must** belong to the class extension of the indicated class description.

I understand this may be more nuanced with OWL 2, but SOSA only refers to OWL 1.

ssn/chapters/ModelPatterns.html
dr-shorthair: > 1. As it stands, that example graph would produce errors when fed to a reasoner. The specification of the [`geo:hasGeometry` property](https://docs.ogc.org/is/22-047r1/22-047r1.html#_01414b7a-8464-4a9b-897d-46c471274ea6) is strict, its domain and ranges are not meant to be implied.

I disagree. As I understand it, the reasoner would not _report errors_, but it would _infer that `earthAtmosphere_StE` is a `geo:Feature`_ ... because of the global `rdfs:domain` constraint. 

The global constraint makes it unnecessary to say `earthAtmosphere_StE a geo:Feature .` since every resource with a `geo:hasGeometry` property is necessarily a `geo:Feature`, which is what the reasoner would infer. 

The function of a reasoner is to add in triples that are implied by the ontology and the data, relieving you of the requirement to include them all in the data explicitly. If any of these triples are inconsistent, we get to dig deeper. In this case, there is nothing inconsistent about `earthAtmosphere_StE a sosa:Sample .` also being inferred to be in the class extension of `geo:Feature`.

ssn/chapters/ModelPatterns.html
dr-shorthair: Thx - I changed to 
> For example [[Lefrancois-et-al-2017]] adapt the SSN Sensing/Sensor/Observation
        pattern to derive explicit Forecasting/Forecaster/Forecast classes.

ssn/rdf/examples/Thermometer.ttl
dr-shorthair: Yes. Removed.

ssn/rdf/examples/Thermometer.ttl
dr-shorthair: `qk:Temperature` is an instance of `qudt:QuantityKind`. 
`qudt:QuantityKind` is effectively a subclass of `sosa:Property`.
  
Here we fix the value of `sosa:observes` to this individual - i.e. we are saying that every `sensor:TemperatureSensor` observes `qk:Temperature`.

ssn/rdf/examples/Thermometer.ttl
dr-shorthair: Removed.

ssn/rdf/examples/InkBird.ttl
dr-shorthair: Fixed

ssn/rdf/examples/InkBird.ttl
dr-shorthair: Conceptually, `qudt:QuantityKind` is a subclass of `sosa:Property`.
I did not state it explicitly here, but in ssn-observation.ttl there is a guarded constraint on the class `sosa:Sensor` that `sosa:observes` takes values from `sosa:Property`. 
Thus, a reasoner can infer that `quantitykind:RelativeHumidity` is a member of `sosa:Property`. 

But I guess it wouldn't do any harm to say this explicitly. 
Done.

ssn/rdf/examples/InkBird.ttl
dr-shorthair: I added into this file
```
qk:RelativeHumidity 
    a sosa:Property ;
    a qudt:QuantityKind ;
. 
qk:Temperature 
    a sosa:Property ;
    a qudt:QuantityKind ;
. 
qudt:QuantityKind
    rdfs:subClassOf sosa:Property ;
.
```

OK?

dr-shorthair:
> @dr-shorthair I would like to clarify the question on `rdfs:domain` (and `rdfs:range`) before giving the green light. OWL defines those as constraints, and [makes it clear](https://www.w3.org/TR/owl-ref/#domain-def):
> 
> > An `rdfs:domain` axiom asserts that the subjects of such property statements **must** belong to the class extension of the indicated class description.

Indeed this is correct: if there are global domain/range constraints associated with a property, then when the property appears in a graph, the subject/object **must** be a member of the domain/range classes. 

However, because of the global constraint in the ontology, this class membership does not need to be asserted explicitly. This is the power of RDFS - a reasoner will infer the class membership. 

OWL adds the ability to define guarded constraints. For example, 'allValuesFrom` constraints say that 'when this property appears *in this context* then the object of the property can be inferred to be a member of the specified class'. 

In both cases, there is no need to assert the class membership explicitly, since it can be inferred from the ontology.

ldesousa:
Alright. I wont block this any further. We can discuss inference some other time.

ldesousa (approved):

ssn/rdf/examples/IBS-TH2-PLUS-brief.ttl
```diff
@@ -0,0 +1,43 @@
...
+
+ex:12gth456a-23190-H a sensor:IBS-TH2-Plus-H ; 
+  sosa:observes <http://qudt.org/vocab/quantitykind/RelativeHumidity> ;
+  system:hasSystemCapability sensor:IBS-TH2-Plus-H-systemCapability ;
+. 
+
+ex:12gth456a-23190-T a sensor:IBS-TH2-Plus-T ; 
+  sosa:observes <http://qudt.org/vocab/quantitykind/Temperature> ;
+  system:hasSystemCapability sensor:IBS-TH2-Plus-T-systemCapability ;
+.
```
sgrellet: @dr-shorthair : 

Seeing the conversation about `qk:Temperature` in the other example files, would that make sense to also point to it in this other ttl (for example consistency) ?
On my side I'm ok keeping this example on qudt but I'm less able to investigate the entailment/reasoner aspects that you are.
Example diversity is also good, but some explanation why there are 2 patterns to describe the same observedProperty could help the reader who won't have the discussion between you and @ldesousa at hand when reading the final spec

sgrellet (approved):
Apart from the small consistency question raised, looks fine to me

ssn/rdf/examples/IBS-TH2-PLUS-brief.ttl
dr-shorthair: I think I fixed this up to be consistent.

Issue #93, opened 2023-09-10
Define scope and output of SHACL validation process


rob-metalinkage:
A process to validate RDF files with SHACL rules has been provided. This fails for mapping files - and does not yet report validation results as artefacts (only action pass or fail).

Consider ideal scope (which files to validate)  and reporting outputs - generate an example report and bring to SOSA meeting with a proposal.

rob-metalinkage:
Need to decide if SHACL rules for enforcing rules in specification text are part of this activity or will be handled by implementations such as https://github.com/opengeospatial/ogcapi-sosa/

dr-shorthair:
Some resources here https://github.com/milieuinfo/RIE-IEPR/tree/main/src/test/resources

rob-metalinkage:
Its clear SHACL is outside the scope of current work.

SHACL rules are included in the https://github.com/opengeospatial/ogcapi-sosa/ and updated to support the latest PWD. 

Consider registering in the SHACL profiles register planned.

PR #256, opened 2024-10-16
239 operating survival ranges, closes #239

### LINKED ISSUE #239: RFC: SurvivalRange / OperatingRange / Range simplification?
(This is part of #230 and comes out from this mornings meeting with @rob-metalinkage and @ldesousa )

Right now, we have OperatingProperty, SurvivalProperty, OperatingRange, SurvivalRange, hasSurvivalProperty, hasOperatingProperty, MeasurementRange. 

untitled

Based on the above: 

**Question 1**: Does it make sense to collapse hasSurvivalProperty / hasOperatingProperty to [ssn:hasProperty](https://www.w3.org/TR/vocab-ssn/#SSNhasProperty) since OperatingProperty and SurvivalProperty already exist?

**Question 2**: Does it make sense to subclass SurvivalProperty from OperatingProperty under the understanding that if the sensor is operating it has survived? @ldesousa made an interesting counter example of systems that go past their operating properties, like the [Voyager Spacecraft](https://en.wikipedia.org/wiki/Voyager_2). My thinking is that this is handled by the difference between the specification of the sensor and the instance of the sensor.

**Question 3**: Can we link SurvivalProperty and OperatingProperty to an sosa:ObservableProperty if only by documenting a design in the narrative, if not ontologically? The reason is that we don't provide a _thing_ to measure in order to determine if the property has been violated. Common sense says that we should look at to the [sosa:observes](https://www.w3.org/TR/vocab-ssn/#SOSAobserves) of the Sensor itself, but it is obvious that some sensors are affected by ancillary properties.

Furthermore:
untitled

**Questions 4:** Does it make sense to create a abstract class Range to do away with OperatingRange, SurvivalRange and MeasurementRange? This is a occurring pattern in the ontology.

oldskeptic:
Implements abstract range class for relevant classes is ssn-system.

I notice that when we talk about a "Survival Range" or "Measurement Range", we don't link it to the Observable Property  that the range should be applied to. Dragging @KathiSchleidt into this.


ssn/chapters/System-capabilities.html
```diff
@@ -909,6 +912,40 @@ <h5><a href="#SSNSYSTEMResolution">ssn-system:Resolution</a></h5>
...
     </section>
 
+    <section class="specterm sosa" id="SSNSYSTEMRange" about="ssn-system:Range" typeof="owl:Class">
+        <h5><a href="#SSNSYSTEMRange">ssn-system:Range</a></h5>
+        <p class="crossreference"><strong>IRI:</strong> http://www.w3.org/ns/ssn/systems/Range</p>
+        <p><strong>an OWL Class</strong></p>
+
+        <em property="rdfs:label">Range</em> -
+        <span property="rdfs:comment skos:definition">A primitive class to represent that the value is within a numerical range.</span>
+        <p>The use of <code>schema:minValue</code> and <code>schema:maxValue</code> are suggested to record the bounds of the numerical range.</p>
```
ldesousa: I rather have the canonical OWL/XSD facets for this. See section 3.3.6: 
https://linked-sdi.com/LinkedSDI.html#literals-facets

ssn/index.html
```diff
@@ -612,6 +612,28 @@ <h3>IP68 Smart Sensor</h3>
...
     </section>
+
+    <section id="InkBird-IBS-TH2">
+      <h3>InkBird IBS TH2 Sensor</h3>
+      <div>
+        <p>These examples describe a commercially available platform, the IBS TH2, marketed under the InkBird brand.</p>
+        
+        <h4>Recording survival and operating ranges for a sensor</h4>
+        
+        <p>In this example we represent the capability of the platform's temperature sensor to measure ambient air and the physical
```
ldesousa: "Temperature sensor of the platform". I heard that "'s" only applies to human beings, but I am not a native.

ssn/index.html
```diff
@@ -612,6 +612,28 @@ <h3>IP68 Smart Sensor</h3>
...
+        survival limits of its plastic construction. Note that this should apply to the platform as well, but in SOSA/SSN survival ranges only
+        apply to sensors and systems, not platforms.</p>
+        <p>An RDF file containing a <a
```
ldesousa, line 553: "A RDF file".

ssn/rdf/examples/InkBird-IBS-TH2-Range.ttl
```diff
@@ -0,0 +1,43 @@
...
+@prefix ssn: <http://www.w3.org/ns/ssn/> .
+@prefix ssn-system: <http://www.w3.org/ns/ssn/systems/> .
+@base <https://rdf.ag/o/equipment/> .
+
+# This example is a partial specification for a common over-the-shelf temperature and humidity sensor.  Different instantions can
+# be derived for each specific physical the product. 
+
+<InkBird-IBS-TH2> a owl:Class;
+    rdfs:label "Inkbird IBS-TH2"@en ;
+    rdfs:subClassOf gs1:Product, sosa:Platform, equipment:Equipment, <https://w3id.org/seas/BluetoothCommunicationDevice
```
ldesousa: The Bluetooth URI here looks incomplete.

ssn/rdf/examples/InkBird-IBS-TH2-Range.ttl
```diff
@@ -0,0 +1,43 @@
...
+    gs1:pip <https://inkbird.com/products/hygrometer-ibs-th2> ;
+    sosa:hosts <IBSTH2TemperatureSensor> .
```
ldesousa, line 24: In my interpretation `sosa:hosts` is not meant to be used this way. It is a Datatype property with `sosa:Platform` as domain  and `sosa:System` as range. But here it is used to relate two classes, instead of instances.

ssn/rdf/examples/InkBird-IBS-TH2-Range.ttl
```diff
@@ -0,0 +1,43 @@
...
+
+<IBSTH2TemperatureSensor> rdfs:subClassOf sosa:Sensor ;
+    rdfs:label "Inkbird IBS-TH2 built-in Temperature Sensor"@en ;
+    sosa:observes <airTemperature> ;
+    ssn-system:hasOperatingRange <IBSTH2TemperatureSensorLimits> ;
+    ssn-system:hasSurvivalRange <IBSTH2SurvivalRange> .
+
+<IBSTH2TemperatureSensorLimits> a ssn-system:OperatingRange ;
+    schema:unitCode qudt-unit:DEG_C ;
+    schema:maxValue "60" ;
```
ldesousa: Minimum and maximum values should be expressed with the W3C infrastructure, in my view. No need to bake in Schema.org (with all that comes with it).

ssn/rdf/examples/InkBird-IBS-TH2-Range.ttl
```diff
@@ -0,0 +1,43 @@
...
+    sosa:hosts <IBSTH2TemperatureSensor> .
+
+<IBSTH2TemperatureSensor> rdfs:subClassOf sosa:Sensor ;
+    rdfs:label "Inkbird IBS-TH2 built-in Temperature Sensor"@en ;
+    sosa:observes <airTemperature> ;
+    ssn-system:hasOperatingRange <IBSTH2TemperatureSensorLimits> ;
+    ssn-system:hasSurvivalRange <IBSTH2SurvivalRange> .
+
+<IBSTH2TemperatureSensorLimits> a ssn-system:OperatingRange ;
+    schema:unitCode qudt-unit:DEG_C ;
```
ldesousa: I am not favourable to this mixing of QUDT with Schema.org. It is much cleaner to simply apply QUDT classes. Here is an example: https://github.com/glosis-ld/glosis/blob/master/glosis_siteplot.ttl#L266

ssn/rdf/examples/InkBird-IBS-TH2-Range.ttl
```diff
@@ -0,0 +1,43 @@
...
+    schema:maxValue "60" ;
+    schema:minValue "-40" ;
+    rdfs:label "Inkbird IBS-TH2 Temperature Sensor Limits"@en .
+
+# Physical limits of the sensor (and platform) where structural failure occurs.
+<IBSTH2SurvivalRange> a ssn-system:SurvivalRange ;
```
ldesousa, line 39: The comments on `IBSTH2TemperatureSensorLimits` apply here too.

ssn/rdf/ontology/extensions/ssn-system.ttl
```diff
@@ -49,6 +49,13 @@ ssn-system: a owl:Ontology , voaf:Vocabulary ;
   vann:preferredNamespaceUri "http://www.w3.org/ns/ssn/systems/" ;
   owl:imports ssn-act: , ssn-obs: , ssn-dep: .
 
+# Range
+
+ssn-system:Range a owl:Class ;
+rdfs:label "Range"@en ;
+skos:definition "A primitive class to represent that the value is within a numerical range."@en ;
+skos:definition "A primitive class to represent that the value is within a numerical range."@en ;
```
ldesousa: Duplicate line.

ldesousa (dismissed):

ssn/rdf/examples/InkBird-IBS-TH2-Range.ttl
oldskeptic: This has to do with this being a "generic" sensor and platform; neither of these should be instances.

ssn/chapters/System-capabilities.html
oldskeptic: I see a problem here in that their implementation lies in using owl:Restrictions. This makes sense in the case of an output value that can never exceed the range, but in this case the actual environmental reading can exceed the range.

oldskeptic:
@ldesousa If we are going to move away from schema.org for values, and I don't have a strong opinion either way, we need to change most of the examples as this is what they are currently using.

ssn/chapters/System-capabilities.html
dr-shorthair: Rob has clarified the requirement here, which is that an instance may need to state a value outside of the range specified in an OWL constraint. So OWL facets don't work here.

oldskeptic: @ldesousa Can I have your 0.02$ on this?

ldesousa: Two questions here:
- `xsd:minInclusive` and `schema:minValue` are semantically equal, therefore I would always opt for the W3C standard.
- `owl:withRestrictions` is more of a design option. I would still prefer to have it, what is the semantic function of a limit that can be surpassed? Perhaps the property itself needs a some more thinking? Two kinds of limits?

ssn/rdf/examples/InkBird-IBS-TH2-Range.ttl
ldesousa: That being the case, what you likely need here is a specialisation of `sosa:hosts`, specifying its particular limitations to domain and range.

ssn/chapters/System-capabilities.html
oldskeptic: @ldesousa The issue lies in the sensor class vs sensor instance problem. 

In _a sensor class_, ```owl:withRestrictions``` make sense in that the sensor is not designed to handle environments outside these values. It's a specification and an ideal model, owl forces the enforcement.

In a _Sensor with Serial Number of a sensor class_, bad things happen to sensor out in the field and out of bounds values happen all the time, be it as magic numbers or through fault.  If we can't list values that are reported by the sensors and out of bounds, it inhibits the capacity of users to ensure data quality.

You are right that xsd:minInclusive and schema:minValue are semantically equal and I have no concerns about using xsd:minInclusive, but feel we should enforce this throughout the specs.

I feel the solution is that the range should list  xsd:minInclusive / xsd:maxInclusive as properties and leave the ontological property enforcement to the user according to the application.  I'll add this to the documentation.

nicholascar: Is not `sh:minInclusive` / `sh:minExclusive` from [SHACL](https://www.w3.org/TR/shacl/#core-components-range) more useful than `xsd:minInclusive`? The SHACL form is also a W3C standard also but one for which there is more modern Semantic Web tooling available.

dr-shorthair: We have not utilised SHACL in this edition. 
The 2017 edition was OWL/RDFS, so we decided to stick with that for this content-focussed update. 

However, I would strongly endorse the development of a SHACL implementation of the SSN Ontology, as a separate standalone document, 'Part 2'. 

Note that https://github.com/w3c/sdw-sosa-ssn/issues/93 mentions some SHACL implementation work undertaken by OGC.

sgrellet: > ... and leave the ontological property enforcement to the user according to the application. I'll add this to the documentation.

That statement is crucial. +1 on this

> SHACL

+1 as well 

> as a separate standalone document, 'Part 2'

@dr-shorthair  : in order to deliver soon would your statement imply that 
- A°/ there should be no xsd:minInclusive awaiting for that Part2 ?
- or B°/ that 1st we should deliver with xsd:minInclusive then update with sh:minInclusive when that Part2 is available ?

dr-shorthair: B. 
Use OWL/RDFS/XSD in this document. 
Provide alternative, but 'equivalent' SHACL implementation in a separate document. 

The SHACL implementation be easier than the work we are doing now, because it will be largely mechanical.

dr-shorthair: And once it is available, the SHACL version is likely to be the most used implementation very quickly ...

oldskeptic:
From meeting, we use ```xsd:minInclusive```. We need documentation for both cases.

ssn/rdf/examples/InkBird-IBS-TH2-Range.ttl
oldskeptic: I'm not sure that this works out really well with the rest of Sosa/ssn. We can keep talking about this in #196 unless you object.

dr-shorthair (approved):

oldskeptic:
@sgrellet @nicholascar @ldesousa Can one of you push the button so we can merge?

ldesousa:
@oldskeptic Sorry, I could not look into this yet. Can it wait for next week?

ldesousa (approved):
To revisit [InkBird-IBS-TH2-Range.ttl](https://github.com/w3c/sdw-sosa-ssn/pull/256/files#diff-38aa1181daac95002a4bd744bb296d424b00a86f406dad327ea3bb5eaed4c889) in a subsequent issue.

PR #264, opened 2024-11-19
253 foi on deployment

### LINKED ISSUE #253: How do we indicate the FoI being Observed by a Sensor?
How do we indicate the FoI being Observed by a Sensor?

If we use an FoI-specific Property (PropertyOfInterest), we can state that the Sensor observes this PropertyOfInterest, that is in turn linked to the FoI.

However, if we use a neutral ObservableProperty, there is no direct way to indicate this link. The current workaround seems to be creating a specialization of Observation for this purpose. If this is a common use case, we may want to add this to SOSA, align with the ObservingCapability class from OMS

dr-shorthair:
Add `hasFeatureOfInterest` to `Deployment`, along with `startTime` and `endTime` - per discussion and agreement in 2024-11-13 telecon; 
removed `forProperty` from `Deployment`; 
clarified documentation around `Deployment`

Closes #253 

Previews: 
https://raw.githack.com/w3c/sdw-sosa-ssn/253-foi-on-deployment/ssn/index.html#Systems-and-their-Deployment-overview
https://raw.githack.com/w3c/sdw-sosa-ssn/253-foi-on-deployment/ssn/index.html#SOSASystem
https://raw.githack.com/w3c/sdw-sosa-ssn/253-foi-on-deployment/ssn/index.html#SOSAhasFeatureOfInterest
https://raw.githack.com/w3c/sdw-sosa-ssn/253-foi-on-deployment/ssn/index.html#SOSAisFeatureOfInterestOf
https://raw.githack.com/w3c/sdw-sosa-ssn/253-foi-on-deployment/ssn/index.html#SOSAendTime


alexrobin:
@dr-shorthair looks good to me.

However, thinking about it, perhaps it's still useful to add `System` to the domain of `hasFeatureOfInterest` as well.

I can see that in the following cases:

1. Somebody may just want to avoid the complexity of describing deployments because the system is always going to be in the same place.

2. Some (sampling) FOIs are intrinsically attached to the system (e.g. a camera frustum, as it is usually defined relative to the camera frame).

dr-shorthair:
Move [this design discussion](https://github.com/w3c/sdw-sosa-ssn/issues/253#issuecomment-2487043123) back to #253

ssn/rdf/ontology/core/sosa-actuation.ttl
```diff
@@ -250,11 +247,14 @@ sosa:startTime
...
   rdfs:isDefinedBy sosa-common: .
 
-sosa:resultTime 
+sosa:endTime 
   schema:domainIncludes sosa:Actuation ;
   schema:domainIncludes sosa:ActuationCollection ;
   rdfs:isDefinedBy sosa-common: .
 
+sosa:resultTime 
+  rdfs:isDefinedBy sosa-common: .
```
ldesousa, line 256: No domain nor range declared for this property. On purpose?

ldesousa (approved):
Just a minor thing to verify.

ssn/rdf/ontology/core/sosa-actuation.ttl
dr-shorthair: We avoided `rdfs:domain` and `rdfs:range` entirely in the core ontology. 
We use `schema:domainIncludes` and `schema:rangeIncludes` in the SOSA modules. 
We use guarded constraints in the SSN modules. 

For `resultTIme`, in sosa-common.ttl see
```
sosa:resultTime 
  a owl:DatatypeProperty ;
  rdfs:label "result time"@en ;
  skos:definition """
  The instant of time when the Execution was completed, i.e. the Actuation or Observation or Sampling  
  """@en ;
  rdfs:comment "The value would usually be encoded using xsd:dateTime xsd:date xsd:gYearMonth xsd:gYear"@en ;
  schema:domainIncludes sosa:Execution ;
```

Since subclassing is not seen by the SOSA modules, I guess we should add matching axioms for the subclasses of Execution (Actuation, Observation, Sampling) and System (Actuator, Sensor, Sampler) in sosa-actuation.ttl, sosa-observation.ttl, sosa-sampling.ttl . 

See https://github.com/w3c/sdw-sosa-ssn/issues/267 to check domain/range and specialization consistency.

rob-metalinkage (approved):

Issue #106, opened 2023-10-25
Distinguish `Property` and `PropertyOfInterest`


maximelefrancois86:
There still is a confusion between whether properties should be specific or generic to features of interest. 

This ambiguity is reported in https://www.w3.org/TR/vocab-ssn/#SSNProperty-instances 

See also footnote 10 in Haller, Armin, et al. "The modular SSN ontology: A joint W3C and OGC standard specifying the semantics of sensors, observations, sampling, and actuation." Semantic Web 10.1 (2019): 9-32. 

> In SSNX, the class of observable properties was defined as a subclass of dul:Quality and as the class of “observable Quality of an Event or Object. That is, not a quality of an abstract entity as is also allowed by DUL’s Quality, but rather an aspect of an entity that is intrinsic to and cannot exist without the entity and is observable by a sensor.” Section 5.3.1.3.4 in [[36]](http://www.w3.org/2005/Incubator/ssn/XGR-ssn-20110628/) further adds that types of properties, such as temperature or pressure should be added as subclasses of oldssn:ObservedProperty instead of individuals. Yet, usage reports [[19]](https://w3c.github.io/sdw/ssn-usage/) of SSNX revealed that most datasets were using observable properties as applicable to many features of interest. See https://www.w3.org/2015/spatial/wiki/What_is_an_instance_of_ssn:Property for more details on this point

Some users of SOSA use `ssn:Property` to classify properties that can be generically used on many features of interest. One advantage is that existing code lists, vocabularies, and taxonomies, may be used. For example:
- the QUDT Quantity Kind vocabulary (ex: [Nozzle Walls Thrust Reaction](http://qudt.org/vocab/quantitykind/NozzleWallsThrustReaction), 
- the British Oceanographic Data Centre Parameter Usage Vocabulary (ex: [Absolute temperature standard deviation of the atmosphere by dry bulb thermometer](http://vocab.nerc.ac.uk/collection/P01/current/CDTSZZ01/)), 
- the ETIM classification standard for technical products (ex: [Tray width in horizontal direction](https://identifier.buildingsmart.org/uri/etim/etim/9.0/prop/EF005872)), etc.

Other users of SOSA use `ssn:Property` to classify properties that are specific to one feature of interest. Below are a few papers and ontologies that use SOSA/SSN with this modeling choice:

[...]

PR #274, opened 2024-12-03
252 remove propertyofinterest

### LINKED ISSUE #212: Extend domains and ranges to PropertyOfInterest
Related to #106 and #126 extend domains and ranges to the class "PropertyOfInterest"

Wherever class "Property", "ObservableProperty" or "ActuatableProperty" is included in the domain or range of an object property, it would be appropriate to also include "PropertyOfInterest"

### LINKED ISSUE #252: Link between PropertyOfInterest and Property
Do we have a way to indicate that a PropertyOfInterest is related to a specific Property? Understand that `Temp of Room3` is `Temperature`

dr-shorthair:
Closes #252 
Closes #212


dr-shorthair:
Preview at https://raw.githack.com/w3c/sdw-sosa-ssn/252-remove-propertyofinterest/ssn/index.html#property-instances

maximelefrancois86:
in the note:  SAREF defines saref:Property as the generic , and saref:PropertyOfInterest as the specific. They are unrelated (no subclass axiom in any direction)

alignment would be: sosa:Property is super class of saref:Property AND saref:PropertyOfInterest

maximelefrancois86 (changes requested):
in the first two paragraphs, it could be appropriate to mention existing usage of sosa:Property that consider it specific vs generic. See beginning of issue #106

in the note: SAREF defines saref:Property as the generic , and saref:PropertyOfInterest as the specific. They are unrelated (no subclass axiom in any direction)

maybe just use SAREF in the example

alignment would be: sosa:Property is super class of saref:Property AND saref:PropertyOfInterest

ssn/chapters/ModelPropertyInstance.html
```diff
@@ -1,17 +1,40 @@
...
+    (e.g. <code>ex:Temperature</code>, <code>ex:OnOffStatus</code>).
+    The catalogs mentioned above primarily support this use case.
+    However, a property may be made specific to a single feature of interest using
+    <a href="#SOSAisPropertyOf"><code>sosa:isPropertyOf</code></a>
+    (e.g. <code>ex:SickChildATemperature</code>, <code>ex:LightStatus</code>).
+
+</p>
+<p>
+    For example, the Property <a href="https://qudt.org/vocab/quantitykind/Temperature"><code>qk:Temperature</code></a>
+    from [[QUDT]] may be specialized to a specific feature of interest, such as the temperature of a particular child or
```
ldesousa: Specialisation primarily refers to classes. Only with ArchiMate have I seen it used between instances, but even there rarely. Perhaps there is a different way to describe this mechanism?

ssn/chapters/ModelPropertyInstance.html
```diff
@@ -1,17 +1,40 @@
 <h3>Generic or Specific Instances of sosa:Property</h3>
 <p>
-    The previous edition of SSN [[vocab-ssn-20171019]] left an ambiguity on whether an instance of <a
-        href="#SOSAProperty"><code>sosa:Property</code></a> should be generic to all features of interest (e.g.,
-    <code>ex:Temperature</code>, <code>ex:OnOffStatus</code>), or specific to a single feature of interest (e.g.,
-    <code>&lt;myBodyTemperature&gt;</code>, <code>&lt;LightStatus&gt;</code>).
-    This version solves the ambiguity by differentiating:
+    The SSN Ontology does not provide a general pattern for describing observable or actuatable properties.
+    It is recommended to use entries from one of the many existing catalogues of properties, such as the <a
```
ldesousa: If you find it relevant: in most cases these properties vocabularies will be expressed with SKOS.

ldesousa (approved):
Just a couple of comments for thought, otherwise good to go.

ssn/chapters/ModelPropertyInstance.html
dr-shorthair: skos:broader/skos:narrower is a standard predicate for generalization/specialization between instances. 
See https://www.w3.org/TR/skos-primer/#sechierarchy 

It is widely recognised that there is crossover between the semantics of skos:Concept and rdfs:Class. 
However, it is also widely recognised that the choice of whether to model a (lower case) concept as a Class or an Individual (e.g. a SKOS Concept) is application specific. Making 'property' a SSN (SOSA) class, and its members as individuals is seen as problematic in some quarters - from the pov of the ISO 19109 General Feature Model having 'Property' as a class already crosses meta-levels uncomfortably. But having 'properties' as individuals does reflect very common usage in observations, so that is what we did. But then when we try to model relationships between properties - particularly specialization/generalization - we inevitably hit the issue under discussion here. 

My general preference is to declare the modeling of property definitions 'out of scope' for SSN, since there is a lot of prior work in the space which it is not necessary to replace. Just use the URIs.

dr-shorthair:
@maximelefrancois86 - I've converted the example to SAREF and generally re-worded to point in that direction. 
Can you please re-review https://raw.githack.com/w3c/sdw-sosa-ssn/252-remove-propertyofinterest/ssn/index.html#property-instances 

Also https://raw.githack.com/w3c/sdw-sosa-ssn/252-remove-propertyofinterest/ssn/index.html#SAREF-alignment - I'm not sure that we have the `hasInput`, `hasOutput`, `hasInputValue` alignments correct yet, also `isPropertyOf` and `isPropertyOfInterestOf`. In particular, do we need the intermediate terms in the http://www.w3.org/ns/sosa/saref# namespace?

ssn/chapters/ModelPropertyInstance.html
ldesousa: OK, but that was not my point, just the wording. I think the two examples are enough to convey these nuances. I just wonder whether using "generalisation" with instances leads to some misunderstanding.

maximelefrancois86 (approved):

maximelefrancois86:
You're right, alignment needs to be reworked. Could you please copy your comments to issue https://github.com/w3c/sdw-sosa-ssn/issues/125 (reopened) ?

PR #278, opened 2025-02-18
Restrictions on object properties in survival ranges example (issue #239)

ldesousa:
Following up on the discussion we had on restrictions to object properties.


dr-shorthair (approved):
Yes, if these sensor types are modeled as classes, then the property values must be modeled as Restrictions, not statements.

oldskeptic:
I agree, this is the way to go.

oldskeptic (approved):

PR #294, opened 2025-03-19
Add (abstract) super-property invForProperty

### LINKED ISSUE #281: bug: union of classes is ok in OWL,  but not union of object properties.
identified here: 

https://github.com/w3c/sdw-sosa-ssn/blob/bb98570b84eb6ae173856dbacc5951071b214dc5/ssn/rdf/ontology/core/sosa-common.ttl#L62-L67

also in the spec.

dr-shorthair:
Closes #281


maximelefrancois86 (approved):

dr-shorthair:
Need one more review on this OWL bugfix please.

ssn/rdf/ontology/core/ssn-common.ttl
```diff
@@ -54,6 +55,10 @@ sosa:hasProperty
 sosa:hasFeatureOfInterest 
   rdfs:isDefinedBy sosa-common: .
 
+sosa:invForProperty
```
ldesousa, line 58: I don't think this is incorrect. However, wouldn't it br more accurate to have this property defined in the [`sosa` module](http://www.w3.org/ns/sosa/)?

dr-shorthair: We use `rdfs:isDefinedBy` to point to the module where the term is introduced. 
`sosa:invForProperty` is introduced in `ssn-common`, and does not appear in any of its dependencies.

rob-metalinkage (approved):

PR #295, opened 2025-03-19
Lefrancois saref alignment

### LINKED ISSUE #125: Add SAREF alignment in Rec doc

dr-shorthair:
Updated to match SAREF new edition

Preview: https://raw.githack.com/w3c/sdw-sosa-ssn/lefrancois-saref-alignment/ssn/index.html#SAREF-alignment 

Closes #125


maximelefrancois86:
This is ready for review. No comment from the ETSI side.
The only modification I see we can discuss, is the ordering of alignments in Section 7.

dr-shorthair:
Happy to promote SAREF above OBOE. 
PROV is more abstract, so the alignment maybe deserves to go first, but I would listen to arguments.

maximelefrancois86:
Please approve

dr-shorthair:
@maximelefrancois86 prepared this in a branch, then @dr-shorthair created the PR. 
We need 2 more reviews to merge.

ssn/rdf/ontology/alignments/sosa-saref.ttl
```diff
@@ -43,112 +43,118 @@ sosa:saref a owl:Ontology ;
...
-  rdfs:subPropertyOf saref:consistsOf .
 
-sosa:Platform
-  rdfs:subClassOf s4syst:System ;
-  rdfs:subClassOf saref:FeatureOfInterest . 
+### Procedures
 
-sosa:hosts
-  rdfs:subPropertyOf s4syst:connectedTo .
+# In SAREF, a `saref:Command` represent the lowest-level directives a device supports and exposes to some network. A saref:CommandOfInterest is a directive actually supported by a device and exposed to some network. A `saref:Operation` is the means of a service to communicate in a procedure-type manner over the network (i.e. transmit data to/from other devices). It is the -machine interpretable- exposure of a -human understandable- command to a network.
```
TallTed:
```suggestion
# In SAREF, a `saref:Command` represent the lowest-level directives a device supports and exposes to some network. A saref:CommandOfInterest is a directive actually supported by a device and exposed to some network. A `saref:Operation` is the means of a service to communicate in a procedure-type manner over the network (i.e., transmit data to/from other devices). It is the -machine interpretable- exposure of a -human understandable- command to a network.
```

ssn/rdf/ontology/alignments/sosa-saref.ttl
```diff
@@ -159,62 +165,236 @@ saref:CommandOfInterest
...
+
+sosa:Platform
+  rdfs:subClassOf s4syst:System ;
+  rdfs:subClassOf saref:FeatureOfInterest . 
+
+
+### Systems
+
+# In SOSA/SSN, `sosa:System` is a generalization of `sosa:Sensor`, `sosa:Actuator` and `sosa:Sampler`, where `sosa:Sensor` covers devices, agents (including humans), or softwares (simulations). 
+# In SAREF, a `saref:Device` is "a tangible object designed to accomplish a particular task. In order to accomplish this task, the device performs one or more functions. An instance of saref:Device represents one specific real world entity.".
```
TallTed:
```suggestion
# In SAREF, a `saref:Device` is "a tangible object designed to accomplish a particular task. To accomplish this task, the device performs one or more functions. An instance of saref:Device represents one specific real world entity.".
```

TallTed (dismissed):
Small fixes

ssn/rdf/ontology/alignments/sosa-saref.ttl
```diff
@@ -153,7 +153,7 @@ saref:hasStateOfInterest
 
 ### Procedures
 
-# In SAREF, a `saref:Command` represent the lowest-level directives a device supports and exposes to some network. A saref:CommandOfInterest is a directive actually supported by a device and exposed to some network. A `saref:Operation` is the means of a service to communicate in a procedure-type manner over the network (i.e. transmit data to/from other devices). It is the -machine interpretable- exposure of a -human understandable- command to a network.
+# In SAREF, a `saref:Command` represent the lowest-level directives a device supports and exposes to some network. A saref:CommandOfInterest is a directive actually supported by a device and exposed to some network. A `saref:Operation` is the means of a service to communicate in a procedure-type manner over the network (i.e., transmit data to/from other devices). It is the -machine interpretable- exposure of a -human understandable- command to a network.
```
TallTed:
```suggestion
# In SAREF, a `saref:Command` represent the lowest-level directives a device supports and exposes to some network. A `saref:CommandOfInterest` is a directive actually supported by a device and exposed to some network. A `saref:Operation` is the means of a service to communicate in a procedure-type manner over the network (i.e., transmit data to/from other devices). It is the -machine interpretable- exposure of a -human understandable- command to a network.
```

ssn/rdf/ontology/alignments/sosa-saref.ttl
```diff
@@ -43,112 +43,118 @@ sosa:saref a owl:Ontology ;
...
-  rdfs:subPropertyOf saref:consistsOf .
 
-sosa:Platform
-  rdfs:subClassOf s4syst:System ;
-  rdfs:subClassOf saref:FeatureOfInterest . 
+### Procedures
 
-sosa:hosts
-  rdfs:subPropertyOf s4syst:connectedTo .
+# In SAREF, a `saref:Command` represent the lowest-level directives a device supports and exposes to some network. A `saref:CommandOfInterest` is a directive actually supported by a device and exposed to some network. A `saref:Operation` is the means of a service to communicate in a procedure-type manner over the network (i.e., transmit data to/from other devices). It is the -machine interpretable- exposure of a -human understandable- command to a network.
```
dr-shorthair: > # In SAREF, a `saref:Command` represents the lowest-level

maximelefrancois86:
The documentation of SAREF V4.1.1 is available here: https://saref.etsi.org/core/v4.1.1/

maximelefrancois86 (approved):

ssn/rdf/ontology/alignments/sosa-saref.ttl
```diff
@@ -43,112 +43,118 @@ sosa:saref a owl:Ontology ;
...
-  rdfs:subPropertyOf saref:isActedUponBy . # may be renamed saref:isTargetOf in V4.1.1
+
+
+### Property
+
+# In SOSA/SSN, a `sosa:Property` can apply to different features of interest. Section 8.8.3 "Feature-specific properties" states that a property may be made specific to a single feature of interest using sosa:isPropertyOf.
+# In SAREF, an instance of `saref:Property` can apply to different features of interest, while an instance of `saref:PropertyOfInterest` is specific to a feature of interest. It is inherent to and cannot exist without that feature of interest.
+# SAREF also defines states, that refer to the identifiable conditions that features of interest are or may be in, and that can be acted upon by devices, such as observed and controlled. States and Properties are parallel, distinct, but not disjoint. Users can chose whichever seems to best fit their needs. 
+# While states can apply to different features of interest, states of interest are specific to a feature of interest.
+# Thus, `saref:Property`, `saref:PropertyOfInterest`, `saref:State`, `saref:StateOfInterest` are all sub-classes of `sosa:Property`.
```
TallTed, line 81: It's important to remember that inline Turtle comments will be lost if the TTL file is ingested by and then regenerated from a triple/quad-store. It may be better to rework these into RDF properties, such as [`skos:prefLabel`](https://www.w3.org/TR/skos-reference/#prefLabel), [`skos:altLabel`](https://www.w3.org/TR/skos-reference/#altLabel), [`skos:definition`](https://www.w3.org/TR/skos-reference/#definition), [`skos:description`](https://www.w3.org/TR/skos-reference/#description), [`rdfs:comment`](https://www.w3.org/TR/rdf-schema/#ch_comment), [`rdfs:label`](https://www.w3.org/TR/rdf-schema/#ch_label), etc.

maximelefrancois86: Sure thing.
But like code formatting and explanatory comments in source code, it may serve a purpose and support code review

dr-shorthair: @maximelefrancois86 is it important for these comments to be preserved if the alignment is loaded?

dr-shorthair:
@maximelefrancois86  - are any changes required because of #317 ? 
In that PR FoI is slightly generalized to also apply to potential executions (through a deployment) as well as actual executions.

ssn/rdf/ontology/alignments/sosa-saref.ttl
```diff
@@ -43,112 +43,118 @@ sosa:saref a owl:Ontology ;
...
+# This part of the file can be read side-by-side with the file sosa-common.ttl 
+
+### FeatureOfInterest
+
+# In SOSA/SSN, `sosa:FeatureOfInterest` is the class of things that are the target of an execution. 
+# In SAREF, `saref:FeatureOfInterest` is the class of real world entities from which a property or a state may be acted upon, such as observed and controlled.
+# So `saref:FeatureOfInterest` is more general than `sosa:FeatureOfInterest`. An entity can be a `saref:FeatureOfInterest` even if no execution has been done about it (yet).
 
 sosa:FeatureOfInterest
   rdfs:subClassOf saref:FeatureOfInterest .
```
sgrellet, line 65: There is a clear difference here in the sense that in SSN/SOSA a feature instance only becomes FoI when acted upon.
Is that really a subClassOf SSN/SOSA is doing and not a restriction ?

ssn/rdf/ontology/alignments/sosa-saref.ttl
sgrellet: to me those comments are crucial so that re-users of the alignment are informed of the rationale behind the alignment and not just blindly using them (will also help maintain this overtime when working on SSN/SOSA revision in couple years :) )

ssn/rdf/ontology/alignments/sosa-saref.ttl
```diff
@@ -43,112 +43,118 @@ sosa:saref a owl:Ontology ;
...
-saref:hasInput
-  rdfs:subPropertyOf <http://www.w3.org/ns/sosa/saref#hasInput> .
+# In SAREF, a device can act upon (OP `saref:actsUpon`) features, properties, or states. SAREF Core defines `saref:observes` and `saref:controls` as sub-properties of `saref:actsUpon`. Property `saref:actsUpon` also applies to functions, commands, and procedure executions.
+# In SOSA/SSN, different object properties are used depending on the context. It would be:
+#
+# - `sosa:hasFeatureOfInterest` from execution to feature of interest
+# - `sosa:hasUltimateFeatureOfInterest` from execution to feature of interest
+# - `sosa:forProperty` from procedure or system to property
+# 
+# Thus, all these properties are sub-properties of the SAREF properties.
```
sgrellet, line 105: this means that we could go from a SSN/SOSA serialisation to a SAREF one with the proposed mapping.

but not vice-versa.

-> is there a additionnal property on SAREF that would allow to do some conditionnal mapping ?

exemple : if saref:actsUpon and property $foo = 'bar' then we are mapping to a sosa:hasFeatureOfInterest (and not the others)

ssn/rdf/ontology/alignments/sosa-saref.ttl
```diff
@@ -159,62 +165,236 @@ saref:CommandOfInterest
...
+
+# SAREF has no equivalent to SOSA/SSN `sosa:Platform` and associated properties:
+#
+# - `sosa:hosts` and its inverse `sosa:isHostedBy`
+# 
+# `saref:FeatureOfInterest` being more general than `sosa:FeatureOfInterest`, this alignemnt maps `sosa:Platform` to `saref:FeatureOfInterest` and `s4syst:System`. 
+
+sosa:Platform
+  rdfs:subClassOf s4syst:System ;
+  rdfs:subClassOf saref:FeatureOfInterest .
```
sgrellet: Regarding "rdfs:subClassOf saref:FeatureOfInterest."

In SOSA, - `A Platform is an entity that hosts other entities, particularly Sensors, Actuators, Samplers, and other `Platforms.
=> It's main reason for existing is to host

Indeed in addition, in some cases in can be a sosa:FoI. 

Looking at SAREF definition of FoI => `"the class of real world entities from which a property or a state may be acted upon, such as observed and controlled."`
=> I feel like it means that we often 'act upon' a sosa:platform. Which to me is not the reality. We often acts upon the other entities hosted by the sosa:Platform

ssn/rdf/ontology/alignments/sosa-saref.ttl
```diff
@@ -159,62 +165,236 @@ saref:CommandOfInterest
...
+# `sosa:System` is positionned in the middle of the hierarchy:
+# `saref:Device` < `sosa:System`  < `s4syst:System`
+#                                 < `saref:FeatureOfInterest` 
+
+saref:Device
+  rdfs:subClassOf sosa:System .
+
+sosa:System
+  rdfs:subClassOf s4syst:System ;
+  rdfs:subClassOf saref:FeatureOfInterest .
```
sgrellet, line 282: sosa:Platform and sosa:System are different classes.

both are proposed to map to `s4syst:System` and 'saref:FeatureOfInterest'. this seems to work in the SOSA -> SAREF direction

is there some extra property that will help differentiate both when going the other way (SAREF -> SOSA) ?

sgrellet (commented):
I gave my 2 cents to the PR. Hope that's useful

Maxime's comments definitively help. But they don't make me a SAREF expert :) Thus my comments are to be taken with a grain of salt.

I guess my main feedback goes back to what I said during webconf 2025-05-07.

I'm not clear on the implications regarding explicitly writing that some SOSA classes are subClassOf SAREF ones.
I could get it when doing SSN/SOSA <-> OMS alignment because we target SSN/SOSA to be the canonical OWL representation of OMS.

Here I guess it's a matter of:
- do we want to document the mapping so that we are able to convert on the fly a SSN/SOSA payload into a SAREF compliant one ? and/or vice-versa ?
- or do we want to strongly relate the ontologies between one another ?  (on this one I may be too much driven by my UML modelling background ^^)

ssn/rdf/ontology/alignments/sosa-saref.ttl
dr-shorthair: I believe this is correct. `sosa:FeatureOfInterest` is a subset of `saref:FeatureOfInterest` - i.e. every individual `sosa:FeatureOfInterest` is a  `saref:FeatureOfInterest` but not vice-versa. 

Even the slight generalization that `sosa:FeatureOfInterest` may be associated with a `Deployment` does not go as far as `saref:FeatureOfInterest` - in SAREF it is every Feature that _has the potential to be the subject of an Execution_, whilst for the SOSA variety it must at least be actively planned through a `Deployment`.

ssn/rdf/ontology/alignments/sosa-saref.ttl
dr-shorthair: Does every `sosa:Platform` have properties that might be actuated or observed? 
If not, then the subsumption relationship to `saref:FeatureOfInterest` is incorrect.

ssn/rdf/ontology/alignments/sosa-saref.ttl
dr-shorthair: Does every `sosa:System` have properties that might be actuated or observed?
If not, then the subsumption relationship to `saref:FeatureOfInterest` is incorrect.

dr-shorthair:
@maximelefrancois86 can you look at the comments from @sgrellet please?

ssn/rdf/ontology/alignments/sosa-saref.ttl
maximelefrancois86: I believe they're better left as comments in the turtle file. If attached to the actual classes and properties, then for each comment about the alignment of two entities, I would have to make an arbitrary choice about on which of these two entities I should attach the comment.

maximelefrancois86:
> @maximelefrancois86 - are any changes required because of #317 ? In that PR FoI is slightly generalized to also apply to potential executions (through a deployment) as well as actual executions.

no functional change, but see commit 101f7a1

maximelefrancois86:
> I gave my 2 cents to the PR. Hope that's useful

Thanks a lot, it is indeed.

> * do we want to document the mapping so that we are able to convert on the fly a SSN/SOSA payload into a SAREF compliant one ? and/or vice-versa ?
> * or do we want to strongly relate the ontologies between one another ?  (on this one I may be too much driven by my UML modelling background ^^)

The primary intention is to be as faithful as possible to the reality.

Given the subClassOf and subPropertyOf axioms are sometimes in one direction and sometimes in the other direction, then it is not really possible to convert on the fly a SOSA/SSN payload to SAREF or vice versa. 

The only good use of this alignment, the one we should promote, is that it allows the **joint** use of SOSA/SSN and SAREF, where  for example users of SAREF would want to benefit from the _ObservationCollection_ class, or the users of SOSA/SSN would want to benefit from things available in SAREF or SAREF extensions.

ssn/rdf/ontology/alignments/sosa-saref.ttl
maximelefrancois86: Yes this is correct

ssn/rdf/ontology/alignments/sosa-saref.ttl
maximelefrancois86: The corollary to the analysis done is:

Going from a SSN/SOSA graph to SAREF and/or vice versa would anyways distort the knowledge embedded in that graph. Each class or property change would imply a loss of knowledge or some possibly untrue addition of knowledge.

I understand "conditional mapping" as the use of equivalentClass/equivalentProperty axioms were one hand is a restriction:

As an example, one could try to formalize differently the alignment between sosa:FeatureOfInterest and saref:FeatureOfInterest as follows:

```
sosa:FeatureOfInterest
  owl:equivalentClass [ 
    a owl:Class ; 
    owl:intersectionOf ( 
      saref:FeatureOfInterest
      [ a owl:Restriction ; owl:onProperty sosa:isFeatureOfInterestOf  ; owl:minCardinality 1 ] 
    )
  ] .  
```

Such conditional alignments would be interesting from a conceptual point of view, but:

1. FeatureOfInterest is an exceptional case. Most pairs of classes couldn't be expressed that way. And I hardly see how such conditional alignments can be defined between properties, given the absence of rich property expression constructors in OWL (apart from inverseOf) 
2. It may be considered too cryptic for the average semantic web enthusiast, even for the members of this group it may make thing difficult to review. I'm afraid it might limit the adoption of this alignment

ssn/rdf/ontology/alignments/sosa-saref.ttl
maximelefrancois86: OK, I delete this axiom in b475dd8

ssn/rdf/ontology/alignments/sosa-saref.ttl
maximelefrancois86: @sgrellet I can't think of a property that could discriminate for the SAREF -> SOSA way, like "a SAREF feature of interest that xxx is a sosa:System"

@dr-shorthair I believe this is correct.

ssn/rdf/ontology/alignments/sosa-saref.ttl
sgrellet: edit : comment moved down the discussion line (I did not see Maxime's overall comment, sorry)

sgrellet:
Was writing this parallel to your contribution

I have one UseCase, slowly arriving in my radar. In EU we have 2 important projects WaterVerse and Water4All which I try to progressively connect.
tl;dr:
- WaterVerse is based on FIWARE, and I understand that at some level in ETSI there is the intent to have [NGSI-LD](https://fr.wikipedia.org/wiki/NGSI-LD) and SAREF 'discuss'
- Water4All is targetting the OGC/W3C paradigm (thus with SSN/SOSA, OMS, SensorThings API as the backbone for the observational aspects)

To me, those mappings/alignments should not only be interesting from a conceptual point of view but really applicable from a data payload one so that 2 information systems built on those 2 paradigms can exchange thanks to this very alignment.

>Given the subClassOf and subPropertyOf axioms are sometimes in one direction and sometimes in the other direction, then it is not really possible to convert on the fly a SOSA/SSN payload to SAREF or vice versa.

"Any chance we could achieve this UseCas ? " was my initial question. 
Ouch, now I'm feeling I'm chasing unicorns here :(

Worth discussing live in a telco

sgrellet:
notes from webconf 2025-05-28 discussion:

Building on Maxime's comment I already quoted above

We are speaking about alignments here, not a converter (which is more the use case I raised).

And the target is not to modify SSN/SOSA &| SAREF so that we can always convert from one to another.

My comment should not block this PR. Accepting it

sgrellet (approved):
approved after webconf 2025-05-28 discussion

PR #299, opened 2025-03-19
better use xsd:decimal than xsd:double

maximelefrancois86:
This is a rule of thumb. See 

> Keil, J. M., & Gänßinger, M. (2022, May). The Problem with XSD Binary Floating Point Datatypes in RDF. In European Semantic Web Conference (pp. 165-182). Cham: Springer International Publishing.


dr-shorthair (approved):
Yes - I agree decimal is better, arguably more correct given it is a textual representation. 
With JM Keil I think we argued the case in one of the units standards threads ;-)

dr-shorthair:
Fixed with 580cd3060259b27866b569bcd45601317ffcf37a

PR #311, opened 2025-04-16
Create battery class and update examples. Closes #310

### LINKED ISSUE #310: Add Battery Class
Add a battery class as an abstract object as per 2025-03-26 meeting.

oldskeptic:
Creates a battery class within system properties as per meeting.

### DIFF
--- a/ssn/chapters/System-capabilities.html
+++ b/ssn/chapters/System-capabilities.html
@@ -1900,4 +1900,42 @@
         </div>
     </section>
 
+    <section class="specterm sosa" id="SSNSYSTEMBattery" about="ssn-system:Battery" typeof="owl:Class">
+        <h5><a href="#SSNSYSTEMBatteryLifetime">ssn-system:Battery</a></h5>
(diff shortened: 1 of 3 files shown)


rob-metalinkage (approved):
Looks OK to me - but I wonder if Battery is overspecific - what if its a fuel cell or a supercapacitor - what about a clockwork mechanism powering a sensor - like a self-winding watch?

oldskeptic:
> Looks OK to me - but I wonder if Battery is overspecific - what if its a fuel cell or a supercapacitor - what about a clockwork mechanism powering a sensor - like a self-winding watch?

The label should probably be "Electrical Battery". The definition of 'an battery' is misused in english/french/german vernacular in that it formally means a "battery of primary or secondary cells" which would cover both fuel cell and a supercapacitor.

As for stored energy sources like spring mechanisms I'm struggling to think an exemplar in the sensor context.

dr-shorthair (approved):

PR #313, opened 2025-04-23
clean up RDFa and toggling of ssn axioms

maximelefrancois86:
It's all as the title says


dr-shorthair (approved):

dr-shorthair:
Thanks for this @maximelefrancois86 - probably should have requested it months ago - it makes the HTML pieces much easier to work with/

KathiSchleidt (approved):

sgrellet (commented):
as agreed in webconf 2025-05-07.
we'll run a HTML validator at the very end to cross-check if some tags appear missing

PR #331, opened 2025-06-13
Rules for consistency of properties in collection members

### LINKED ISSUE #328: Clarify collections consistency rules
See https://github.com/w3c/sdw-sosa-ssn/pull/326#pullrequestreview-2896057083 

Provide examples to illustrate each rule

dr-shorthair:
Preview: https://raw.githack.com/w3c/sdw-sosa-ssn/328-collections-consistency-rules-examples/ssn/index.html#ModelCollectionRules 

Closes #328


rob-metalinkage (approved):
quite a lot of changes - but the patterns look good and comprehensive.  Will try to validate examples again and generate a new PR to fix if needed,

maximelefrancois86 (approved):
in general: fine with the proposed changes.  It could be nice to have a little bit more explanations in the examples directly (for example as Turtle comments), for example: OCrule2.ttl has two observations. one explicits the same observedProperty as the collection, so this is compliant with the rule. The second observation has the observedProperty implicit. Rule 2 implies that the observedProperty can be inferred.

oldskeptic (approved):

rob-metalinkage:
phenomenontime rangeIncludes a time TemporalEntity - these seems asymmetric with simple scalars for resultTime - should examples clarify options - they lead to one interpretation which may be at odds with valid and likely  implementation choices.

dr-shorthair:
> phenomenontime rangeIncludes a time TemporalEntity - these seems asymmetric with simple scalars for resultTime - should examples clarify options - they lead to one interpretation which may be at odds with valid and likely implementation choices.

'Twas ever thus. See 2017 version. 
- `resultTime` is a Datatype Property https://www.w3.org/TR/vocab-ssn/#SOSAresultTime 
- `phenomenonTime` is an Object property https://www.w3.org/TR/vocab-ssn/#SOSAphenomenonTime

Remember: `phenomenonTime` relates to the world time, while `resultTime` relates to the data time. 
The examples are consistent - they only show `phenomenonTime` cases on both the collection and members. 
In general, `phenomenonTime` is usually the more important property from the application viewpoint. 

What is _not_ shown is an example showing a _range_ of resultTimes on a collection. 
I'm not sure how to do that. A list is fine - just repeat the `resultTime` property. But a range is not syntactically supported by any XSD datatype.

PR #426, opened 2026-03-25
Initial D28, D29 & D30 implementations

nicholascar:
* updated info about D28, D29 & D30
* indicated a few SOSA elements implemented by D28, D29 & D30
    * more to come later


dr-shorthair (approved):
Thanks Nick - just what we were looking for.

PR #458, opened 2026-06-03
Markup fixes

bert-github:
It seems HTML5 requires the children of a DL to be either DIV elements or DT & DD elements, but not a mixture of DIV, DT and DD elements. So, whenever a DL had a DIV as child, I wrapped all other DT and DD elements in DIVs.

Also, it seems it is not allowed to have a DT after a DD inside a DIV, so I split some DIVs into two DIVS just before the DT.


ssn/chapters/DUL-alignment.html
```diff
@@ -43,58 +47,70 @@ <h4>Class Alignments</h4>
...
...valanche basic participants, or if we also want to add water, human agents, etc., that makes the identity of an avalanche change. Anyway, this approach to event classification is based on the designer's choices, and more accurately mirrors lexical or commonsense classifications (see. e.g. WordNet 'supersenses' for verb synsets).  Ultimately, this discussion has no end, because realists will keep defending the idea that events in reality are not changed by the way we describe them, while constructivists will keep defending the idea that, whatever 'true reality' is about, it can't be modelled without the theoretical burden of how we observe and describe it.  Both positions are in principle valid, but, if taken too radically, they focus on issues that are only partly relevant to the aim of computational ontologies, which only attempt to assist domain experts in representing what they want to conceptualize a certain portion of reality according to their own ideas.  For this reason, in this ontology both events and situations are allowed, together with descriptions, in order to encode the modelling needs, independently from the position (if any) chosen by the designer.">dul:Event</span>
```
TallTed, line 51: Are these *dozens* of lines really the "title" of this `<span>`?

TallTed: Merged without answer or comment? That's troubling.

PR #459, opened 2026-06-19
Small change to the definitions of System and Procedure

maximelefrancois86:
This PR proposes a small clarification of the definitions of `System` and `Procedure`, with particular attention to the interpretation of software-based systems.

It originates from a collaboration with a researcher at LUT University (Finland). In this context, a concrete Python script was initially intended to be modeled as a System (a “Calculation”), while the abstract notion of a “Task” that the script is meant to carry out (e.g., fatigue calculation) was modeled as a Procedure.

While this interpretation is understandable given the current definitions, I consider that the existing wording is not sufficiently precise and allows for ambiguity between systems as artifacts that perform computation and procedures as abstract specifications of computational or observational behaviour.

In my view, a Python script, like other software programs, is more appropriately modeled as a Procedure, since it encodes a computational method or algorithmic specification. The relationship between a concrete computational artifact (e.g., a script) and the abstract task it operationalizes can then be captured outside SSN/SOSA, for example using a UML-style notion such as _manifestation_, which relates an artifact to the specification it embodies.

Accordingly, this PR makes the following minimal changes:

- In the definition of `sosa:System` and its subclasses, it replaces “software (simulation)” with “or a software-based system, including systems performing simulations”
- In the definition of `sosa:Procedure` and its subclasses, it replaces “workflow, protocol, plan, algorithm, computational method” with “workflow, protocol, plan, algorithm, **program**, or computational method”

These changes aim to improve conceptual clarity.

### DIFF
--- a/ssn/chapters/Actuation.html
+++ b/ssn/chapters/Actuation.html
@@ -51,7 +51,7 @@
         </p>
         <p>
             <em>Actuating Procedure</em> &mdash;
-            workflow, protocol, plan, algorithm, or computational method specifying how to make an Actuation
+            workflow, protocol, plan, algorithm, program, or computational method specifying how to make an Actuation
         </p>
         <p>
             An Actuating Procedure is re-usable, and might be applied in many Actuations.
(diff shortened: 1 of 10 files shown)


dr-shorthair (approved):

dr-shorthair:
@rob-metalinkage @ldesousa @sgrellet @kjano could you review this please

ldesousa (approved):

PR #525, opened 2026-09-12
Repository checks in Linked-Data Python, and a script that publishes the namespace

maximelefrancois86:
Two things this branch does, both about keeping the repository and what it publishes in agreement.

## The checks, rewritten in Linked-Data Python

`ssn/scripts/check_repository.ldpy` is written in [Linked-Data Python](https://linked-data-python.readthedocs.io), a superset of Python whose syntax includes Turtle, so an RDF query reads as the pattern it matches rather than as a chain of rdflib calls:

```ldpy
for term, ontology in m{ ?term rdfs:isDefinedBy ?ontology }:
```

On top of what it already verified — unused chapters and images, dangling internal links, Turtle that does not parse, terms defined twice or never, labels, domains and ranges, deprecation, documentation anchors, prefixes, examples — it now **compares the Turtle definitions with the prose** of the specification, and says which side of a divergence was edited last.

## A script that publishes the namespace

`ssn/scripts/publish_namespace.py` does the two mechanical halves of publishing SOSA/SSN under `https://www.w3.org/ns/`, into a checkout of [w3c/ns](https://github.com/w3c/ns). Neither half is decided by a hand-written table; both come from what the files and the specification declare.

**The files.** Each ontology is copied to the place its IRI names, with the RDF/XML alternative beside it:

| ontology IRI | destination |
|---|---|
| `http://www.w3.org/ns/sosa/act/` | `sosa/2023/act/sosa-act.ttl` + `.rdf` + an HTML stub |
| `http://www.w3.org/ns/sosa/oboe` | `sosa/2023/oboe.ttl` + `.rdf` |
| `http://www.w3.org/ns/sosa/system-capability-properties#` | `sosa/2023/system-capability-properties.ttl` + `.rdf` |

An `owl:versionIRI` wins over that derivation. The RDF/XML is written deterministically: rdflib serialises in the hash-seed order of the interpreter, so the script canonicalises the graph, sorts the triples, binds the prefixes in a fixed order and re-execs itself with `PYTHONHASHSEED=0`. Two runs produce the same bytes, and a namespace tree under version control does not churn.

**The rules.** Into a delimited block of each namespace's `.htaccess`, the script writes the rewrite rules that serve the terms: to the anchor of the specification for a client that wants HTML, to the version IRI of the module that defines the term for everyone else. The anchors come from this repository, where every term is marked up as

```html
<section class="specterm sosa" id="SOSAActuator">
    <h5><dfn>sosa:Actuator</dfn></h5>
    <p class="crossreference"><strong>IRI:</strong> http://www.w3.org/ns/sosa/Actuator
```

and the terms of the 2017 edition are read from the namespace tree itself, so that each term IRI serves the most recent edition that declares it. A term IRI no edition declares matches no rule, and the server answers 404 instead of redirecting to an invented fragment.

`check_repository.ldpy` gains `check_publication()` for the same purpose: the `=== Ontology Declarations ===` listing, which the rewrite had dropped, and two non-fatal checks — every declared ontology needs a home under `/ns/sosa/` or `/ns/ssn/` (`sosa-sdo.ttl` has none: it declares `https://example.org/sosa-sdo-mapping`), and a version IRI must be its ontology IRI with one edition inserted.

## Where this is used

The namespace side is being prepared in `w3c/ns`, where both the 2017 and the 2023 editions will be served side by side. Running the script there places 25 ontologies and writes the rules for 29 modules and 198 terms; the regression tests of both namespaces pass against a local Apache.

## Note for reviewers

`check_repository.ldpy` currently fails three checks on this branch, all of them about the specification rather than about this tooling: dangling internal links, terms used but never defined, and terms missing a label or a definition.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dr-shorthair (approved):

dr-shorthair:
AFAICT the build failure will be fixed by the other PRs already stacked up, 
in particular the OWL DL fixes

ldesousa (approved):

PR #529, opened 2026-09-12
Declare what OWL 2 DL asks for, and fix two plain slips

maximelefrancois86:
Takes the part of #528 that needs no decision. **1055 OWL 2 DL violations down to 86**, and fifteen of the eighteen core, extension and vocabulary modules are now in the profile.

Nothing an ontology already asserted changes, with one exception noted at the end.

## Three commits

**1. Declare the vocabularies each module annotates with** — 969 violations

OWL 2 DL asks that every entity an ontology uses be declared somewhere in its imports closure, and the metadata vocabularies the modules are annotated with never were. Each module now declares what it uses, right after its ontology header:

```turtle
## Terms of other vocabularies, declared here because OWL 2 DL asks that
## every entity an ontology uses be declared in its imports closure

dcterms:created a owl:AnnotationProperty .
dcterms:creator a owl:AnnotationProperty .
...
foaf:name a owl:AnnotationProperty .
foaf:Agent a owl:Class .
```

Per module, not in one shared block up the imports chain: the file then says what it needs whichever way the import graph is rearranged later. Repetitive, but each declaration is a single axiom that costs nothing to repeat.

That is #528 §1, §2 and §4 — including `foaf:Agent`, `voaf:Vocabulary` and `skos:Concept`. Declaring `foaf:Agent` also settles the *non-absolute IRI* report of §2: once the class is declared, the anonymous `dcterms:creator` node is a legal annotation value and the OWL API stops complaining about it. **Replacing that anonymous node with an IRI for the working group is still open** — it is the other half of §2, and it changes what the file asserts, so it is left out here.

**2. Declare `sosa:Battery` a class** — 1 violation

It only ever appeared as the subject of `rdfs:subClassOf`, so nothing said it was a class.

**3. Restrict `sosa-oms:preparedSample`, the property that exists** — 16 violations, and **fixes #517**

```diff
 sosa-oms:PreparationStep
-  rdfs:subClassOf [ a owl:Restriction ; owl:onProperty sosa:preparedSample ; owl:cardinality "1"^^xsd:nonNegativeInteger ] ;
-  rdfs:subClassOf [ a owl:Restriction ; owl:onProperty sosa:preparedSample ; owl:allValuesFrom sosa:Sample ] ;
+  rdfs:subClassOf [ a owl:Restriction ; owl:onProperty sosa-oms:preparedSample ; owl:cardinality "1"^^xsd:nonNegativeInteger ] ;
[...]

Issue #528, opened 2026-09-12
No core, extension or vocabulary module is in the OWL 2 DL profile


maximelefrancois86:
The repository check now asks the OWL API whether each module — **with its imports** — is in the **OWL 2 DL** profile, and asks HermiT whether any class is unsatisfiable.

```sh
pip install linked-data-python owlapy   # and a Java runtime, for the OWL API and HermiT
python -m ldpy ssn/scripts/check_repository.ldpy
```

The reasoning side is clean: all 18 modules are consistent, and every class can have an instance. The profile side is not. **Every one of the 18 core, extension and vocabulary modules is outside OWL 2 DL**, on 1055 violations between them.

That matters because OWL 2 DL is what keeps reasoning decidable, and what the reasoners the specification expects are built for. It is a syntactic condition, so losing it costs nothing visible — the Turtle still parses, the documentation still renders — until someone runs a reasoner that refuses the input, or silently drops an axiom (see §6 below, where one already is).

A violation is counted once, against the module that holds the offending axiom, even though it takes every importing module out of the profile too.

| Module | Violations |
|---|---|
| `ssn/rdf/ontology/core/sosa-common.ttl` | 212 |
| `ssn/rdf/ontology/extensions/sosa-oms.ttl` | 174 |
| `ssn/rdf/ontology/core/sosa-sampling.ttl` | 143 |
| `ssn/rdf/ontology/core/sosa-observation.ttl` | 138 |
| `ssn/rdf/ontology/core/sosa-actuation.ttl` | 93 |
| `ssn/rdf/ontology/extensions/sosa-system.ttl` | 40 |
| `ssn/rdf/ontology/core/ssn-deprecated.ttl` | 40 |
| `ssn/rdf/vocabularies/system-capabilities-properties.ttl` | 39 |
| `ssn/rdf/ontology/core/sosa.ttl` | 36 |
| `ssn/rdf/vocabularies/system-environment-properties.ttl` | 30 |
| `ssn/rdf/ontology/extensions/sample-relations.ttl` | 27 |
| `ssn/rdf/ontology/core/sosa-deprecated.ttl` | 18 |
| `ssn/rdf/ontology/core/ssn.ttl` | 13 |
| `ssn/rdf/ontology/extensions/ssn-oms.ttl` | 13 |
| `ssn/rdf/ontology/core/ssn-actuation.ttl` | 11 |
| `ssn/rdf/ontology/core/ssn-common.ttl` | 10 |
[...]

Issue #280, opened 2025-02-26
Divergence between Sensor (Device, agent (including humans), or software (simulation)) and Actuator (device)


KathiSchleidt:
The definition of Sensor is quite wide, `Device, agent (including humans), or software (simulation)` involved in or implementing a procedure

When we compare this with Actuator, it is much narrower, just a `device` that is used by, or implements, an (Actuation) Procedure 

Looking at the underlying RDF, there's nothing to back up this difference. Thus, shouldn't we expand the textual definition of Actuator to be parallel with Sensor? Support human actuators?

alexrobin:
@KathiSchleidt I fully agree with this.

dr-shorthair:
Yes - there are many inconsistencies and asymmetries in the text. 
It'll require a close reading to tease all these out as we go public. 

I've added `documentation` and `editorial` tags so we can gather them all together.

dr-shorthair:
@maximelefrancois86 created a related issue #284 with the following comment: 

> a sosa:System is "a piece of infrastructure that implements a Procedure."
> System generalizes the Actuator, Sensor, and Sampler classes.
> In its definition, sosa:Sensor covers devices, agents (including humans), or softwares (simulations).
> Therefore, agents (including humans), or softwares (simulations) are pieces of infrastructures.
> Maybe we should relax the definition of sosa:System to "an entity (such as a piece of infrastructure) that implements a Procedure."

I'll close that and we'll manage it all together here.

PR #530, opened 2026-09-12
Bring the SAREF alignment back in step with SOSA and with SAREF V4.1.1

maximelefrancois86:
The SAREF alignment was last revised in May 2025. Since then SOSA grew a `sosa:Asset` class above `Platform` and `System`, moved a batch of sub-class and sub-property axioms out of SSN into the `sosa-*` modules, and reworded a number of definitions — while SAREF published a V4.1.1 ontology that does not quite match its own Technical Specification. This brings the file back in step with both.

Nothing about the *intent* of the alignment changes. What changes is which terms it names, where it says things, and what its comments claim.

## 1. Follow the SAREF ontology where it and TS 103 264 disagree

The file imports `<https://saref.etsi.org/core/v4.1.1>`, and that ontology and **ETSI TS 103 264 V4.1.1** disagree on three terms:

| | TS 103 264 V4.1.1 | ontology at `core/v4.1.1/saref.ttl` |
|---|---|---|
| clause 5.2, 5.11.1 | `saref:actsUpon` | **`saref:targets`** |
| clause 5.11.1 | `saref:isActedUponBy` | **`saref:isTargetOf`** |
| clause 5.11.1 | `saref:hasStartTime` (DP) | **not declared** |

The renames are recorded only in `skos:historyNote` on the individual terms — neither the TS change list nor the ontology's own `rdfs:comment` change list mentions them.

A reasoner sees the ontology, not the TS, so the alignment now follows the ontology. Five axioms were naming properties that do not exist. The `sosa:startTime` equivalence was worse than dead: with no declaration for `saref:hasStartTime`, an OWL parser cannot decide whether the equivalence holds between object or between data properties, and **drops the axiom entirely** — verified against the OWL API. The alignment is kept and the declaration made here, until V4.1.2 settles all three. A note at the top of the file records why.

The rename also makes the file do what its comments always said. `saref:isTargetOf` is declared `owl:inverseOf saref:targets`, where the old undeclared names carried no inverse — so `sosa:isFeatureOfInterestOf` and `sosa:isUltimateFeatureOfInterestOf` are now genuinely covered, exactly as the `# this also covers …` comments promise.

## 2. Say once, on `sosa:Asset`, what Platform and System inherit

`sosa:Asset` arrived in `sosa-common` in January, above `sosa:Platform` and `sosa:System`, and the alignment gained a line for it without the two older lines being revisited. Both had become entailed:

```diff
 sosa:Asset
-  rdfs:subClassOf s4syst:System .
-
-sosa:Platform
-  rdfs:subClassOf s4syst:System .
-
-sosa:System
-  rdfs:subClassOf s4syst:System ;
-  rdfs:subClassOf saref:FeatureOfInterest .
+  rdfs:subClassOf s4syst:System ;
+  rdfs:subClassOf saref:FeatureOfInterest .
```

The comment on `Platform` claimed this alignment maps it to `saref:FeatureOfInterest` *and* `s4syst:System`, and no axiom ever did the first half — HermiT confirms `sosa:Platform` was not a `saref:FeatureOfInterest`. Carrying it on `Asset` makes the comment true and covers `System` by the same stroke.

## 3. Refresh the SOSA/SSN definitions the comments quote

The comments carry the reasoning behind each axiom, and they quote definitions that have moved. Four had gone from stale to wrong:

- three say `sosa:madeBySensor`, `madeByActuator` and `madeBySampler` are sub-properties of `sosa:madeBySystem` **"in SSN, not in SOSA"**, and a fourth says the same of `sosa:Sampling` under `sosa:Procedure`. Those axioms moved into the `sosa-*` modules in October 2025, and `sosa:Sampling` is under `sosa:Execution`, not `sosa:Procedure`;
- `sosa:propertyFor` no longer reads "A relation between a Property and some entity", and its third example named a `SystemCapability` — a class that no longer exists anywhere in the ontology;
- `sosa:Actuator` "covers devices", with a note that [#280](https://github.com/w3c/sdw-sosa-ssn/issues/280) was discussing whether to relax it. #280 was settled, and in the direction it asked: an Actuator may be a device, a piece of infrastructure, an agent, or a software-based system. That is what made `saref:Actuator` the more specific of the two all along.

`FeatureOfInterest`, `Property`, `hasProcedure`, `Sensor` and `System` are refreshed too, and the hard clause number for "Feature-specific properties" is replaced by its title, section numbers being generated.

## What was deliberately left alone

An entailment check — each alignment axiom removed in turn, then asked of HermiT — finds **twelve** axioms the imports already entail, not the two this PR removes. The other ten are kept:

- the four above, whose comments now say they are entailed. They were put there on purpose, "in case it is used only with SOSA (without SSN)"; that reason has expired, but an alignment file is read as much as it is reasoned over;
- `sosa:MaterialSample`, `SpatialSample`, `StatisticalSample` under `saref:FeatureOfInterest`, entailed through `sosa:Sample`;
- `sosa:hasProcedure ⊑ saref:isTargetOf`, entailed through `sosa:propertyFor`;
- `sosa:madeBySystem ≡ saref:madeBy` and `sosa:madeExecution ≡ saref:madeExecution`, which entail **each other** through the two `owl:inverseOf` declarations — dropping either is safe, dropping both loses the alignment.

Say the word and they go in a follow-up.

Three OWL 2 DL profile violations also remain in this file — `dcterms:rights`, `foaf:name` and `foaf:Agent` undeclared. They are the same family as #528 §1 and §2, which #529 fixes for the core, extension and vocabulary modules; the alignments were out of scope there.

## Verification

Loaded with SAREF Core V4.1.1 and SAREF4SYST V1.1.2 on disk (V1.1.2 is the latest ETSI publishes), through the OWL API and HermiT:

| | before | after |
|---|---|---|
| OWL 2 DL violations held by this file | 5 | 0 (bar the three above) |
| consistent | yes | yes |
| unsatisfiable classes | 0 | 0 |
| same, with all of SSN loaded alongside | 0 unsat | 0 unsat |

Every SOSA/SSN ↔ SAREF subsumption the alignment entails was enumerated before and after. **Four are gained** — `sosa:Asset` and `sosa:Platform` under `saref:FeatureOfInterest`, `sosa:isFeatureOfInterestOf` and `sosa:isUltimateFeatureOfInterestOf` under `saref:isTargetOf`. **Four disappear**, all of them bridges to `saref:actsUpon` and `saref:isActedUponBy`, which exist only because this file used to conjure them; the real SAREF hierarchy was always reached through `saref:observes` and `saref:controls`. Everything else is identical, term for term.

`check_repository.ldpy` reports no change in any other section.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ssn/rdf/ontology/alignments/sosa-saref.ttl
+++ b/ssn/rdf/ontology/alignments/sosa-saref.ttl
@@ -267,10 +267,12 @@ sosa:startTime owl:equivalentProperty saref:hasStartTime .
 #
 # - `sosa:hasDeployment` and its inverse `sosa:deployedAsset`
 # - `sosa:isHostedBy` and its inverse `sosa:hosts`
-# 
+#
+# `sosa:Asset` is the resource involved in a Deployment, and `sosa:Platform` and `sosa:System` are both subclasses of it. It is the one place to say that these are systems in the SAREF4SYST sense, and features of interest in the SAREF sense — `saref:FeatureOfInterest` being more general than `sosa:FeatureOfInterest`, an entity can be one without being the target of any execution.
 
 sosa:Asset
-  rdfs:subClassOf s4syst:System . 
+  rdfs:subClassOf s4syst:System ;
+  rdfs:subClassOf saref:FeatureOfInterest . 
 
 
 ### Deployment
@@ -288,10 +290,7 @@ sosa:Asset
 #
 # - `sosa:hosts` and its inverse `sosa:isHostedBy`
 # 
-# `saref:FeatureOfInterest` being more general than `sosa:FeatureOfInterest`, this alignemnt maps `sosa:Platform` to `saref:FeatureOfInterest` and `s4syst:System`. 
-
-sosa:Platform
-  rdfs:subClassOf s4syst:System . 
+# `sosa:Platform` is a subclass of `sosa:Asset`, so it is already a `s4syst:System` and a `saref:FeatureOfInterest`. 
 
 
 ### Systems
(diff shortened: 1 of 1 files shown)


dr-shorthair (approved):
Good to have this updated

rob-metalinkage (approved):

